PR #3, opened 2023-04-03
Feat : drawer-screen improvements

surbhidavara99:


client/android/gradle.properties.example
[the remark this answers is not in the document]
surbhidavara99: I have renamed it but it's still showing in the deleted format

client/screens/NavigationMobile.js
andrew-bierman, line 1: Is signOut working on your end? Might need to add that into revised version

client/app/drawer/index.js
andrew-bierman, line 1: Should this be in component folder instead of app? Just for organization

surbhidavara99: We are making it full-screen so we need navigation to redirect it. That's why we can't move into a component folder

client/screens/NavigationMobile.js
surbhidavara99: Yes, It's fixed. Please check now

PR #52, opened 2023-05-26
Trip create

mdajminpradhan:


andrew-bierman:
This looks good, but my only concern is changing the trip model start and end dates to type String instead of keeping them as dates. Eventually, we will want to perform date logic on them, such as seeing if our weather object does not fall within in the selected dates (so we can exclude it). 

What do you think ?

PR #62, opened 2023-06-01
Map box fixes

BilalSadiq421:


andrew-bierman:
So this merge is just adding the issue templates is that correct ?

PR #66, opened 2023-06-02
Gulzar dev to testing, trip details page

andrew-bierman:


andrew-bierman:
need to integrate cords logic into existing shape logic to differentiate between a trail and location

PR #83, opened 2023-06-26
inegrate cors security policy

aezazali1997:
added cors functionality so that only a specific origin can make requests to backend


andrew-bierman:
Based on follow up "Mobile apps do go use cors. they have some different mechanism. only web goes through some specific port. so cors exists only on web."

Lgtm

PR #84, opened 2023-06-26
frontend validation along with backend

aezazali1997:
Added frontend validation with validator library and made some changes to backend email validation


client/screens/LoginScreen.js
andrew-bierman, line 1: So this looks fine, i'm leaving this comment as a note to myself (you don't need to do this). We have a ton of inputs on this app. We should create a more dynamic approach to validating these. Probably a Form or error class or something that can handle error more dynamic so we don't need to do so many state variables for errors.

client/store/authStore.js
andrew-bierman, line 1: Alright i know some of this logic was here before, not your fault. But why are we doing UI Toasts in our redux store ? 

Note to myself to come fix this

server/controllers/userController.js
andrew-bierman, line 1: Why remove the generate Auth token ?

andrew-bierman (changes requested):
You did a good job, there were some strange things that were in these files that I missed before. Going to want to clean these up before merging this one in

client/screens/LoginScreen.js
aezazali1997: I have a suggestion. how about we move to Formik. so we can handle it more dynamically. Formik seems to be a good approach.

PR #108, opened 2023-07-04
Fix/refactoring pack table

andrew-bierman:


andrew-bierman:
@aezazali1997 see my modifications here. I think it would be nice if we could have buttons in each category section, that opened a modified modal. Example, Food section header would have a button to Add Food Item. This modal would not have the option to select a different category, we would just have that preset before sending to backend.

PR #124, opened 2023-07-09
Mikibo refactor map buttons

mikib0:


andrew-bierman:
nice work, i'm getting a couple of errors trying to use the buttons such as the location, switching through styles.

andrew-bierman:
It looks like preview map zoom is not showing trail the way it was previously.
Screenshot 2023-07-09 at 6 03 51 PM

client/components/map/MapButtonsOverlay.js
andrew-bierman, line 1: On iOS devices with notch, top buttons are too far in top. MIght need to do a SafeAreaView to stop that

client/components/map/NativeMap.native.js
andrew-bierman, line 1: Preview map zoom not properly fitting trail anymore

client/components/map/MapButtonsOverlay.js
andrew-bierman, line 1: Mapbox errors on location button, changing through styles in full screen mode

andrew-bierman (changes requested):
Looks good overall, just need to fix up a couple functionality things / errors and this is good to merge. See comments when possible please.

mikib0:
> 

Ah, I see. But switching styles work right, eventhough some errors get thrown. I also now get this "No location provider available" error when I click the location button. This is odd, everything was working w/o errors yesterday 😬😀. But it is alright, I'll check them.

client/components/map/NativeMap.native.js
mikib0: Fixed👍

client/components/map/MapButtonsOverlay.js
mikib0: I wrapped the component in a SafeAreaView; Please check again if the problem is resolved.

mikib0: For the location problem...is the device's location turned on? The functionality worked and I stopped getting the error after I turned on my phone's location.

mikib0: The errors on changing styles seem go away too after turning on the location.

client/utils/mapFunctions.js
```diff
@@ -1,6 +1,7 @@
 import * as Location from "expo-location";
 import * as FileSystem from 'expo-file-system';
-// import { saveAs } from 'file-saver';
+import { saveAs } from 'file-saver';
```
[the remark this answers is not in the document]
mikib0: Yea we can just simply implement the logic differently for each platform.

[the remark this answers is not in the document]
andrew-bierman: Yes, I went ahead and added some custom imports to address this based on platform

client/components/map/MapButtonsOverlay.js
andrew-bierman: Can we prompt for the location with that expo location library ? I think we can, that might improve the UX

andrew-bierman: iOS throwing this on location press, with location permissions enabled 

` ERROR  Mapbox error CameraUpdateItem.execute  paramError("latitude should be between -90 and 90 but it was -122.406417") {"level": "error", "message": "CameraUpdateItem.execute  paramError(\"latitude should be between -90 and 90 but it was -122.406417\")"}
› Reloading apps`

andrew-bierman: Seems to still persist and some iOS cuts off the buttons entirely. Might need to do something with the StatusBar from react native and maybe pass that value in with some padding like marginTop: StatusBar.currentHeight + 20. Just an idea, not sure if this would work

andrew-bierman:
On web, we had to remove the style layers before adding new ones. Might need to see if something like that can be done with RNMapbox

mikib0:
> On web, we had to remove the style layers before adding new ones. Might need to see if something like that can be done with RNMapbox

I don't think it's necessary to do something like that on mobile, because we are rerendering the component when the style changes. Or what do you think?

client/components/map/MapButtonsOverlay.js
mikib0: > Can we prompt for the location with that expo location library ? I think we can, that might improve the UX

Right, I'll look into this.

>  ERROR  Mapbox error CameraUpdateItem.execute  paramError("latitude should be between -90 and 90 but it was -122.406417") {"level": "error", "message": "CameraUpdateItem.execute  paramError(\"latitude should be between -90 and 90 but it was -122.406417\")"} › Reloading apps

This is strange. Please check (maybe the `custom location` config) if the latitude value is within the valid range.

andrew-bierman: On android do you not get that CameraUpdateItem error ?

mikib0: No, I don't.

andrew-bierman:
I'm going to merge this in to keep the andrew_testing branch up to speed, but if you could review my feedback and make sure we covered everything that would be great @mikib0

andrew-bierman (approved):
lgtm

PR #136, opened 2023-07-13
Virtualized list bug fix and css issues resolved

umerbaigh:


andrew-bierman (changes requested):
Make a brand new branch from andrew_testing and roll and delete this one off the git. Not sure what happened here

PR #140, opened 2023-07-17
Copy public pack

aezazali1997:
added copying icon that is pressable and some redux stats and some backend routes and controllers for handling of duplicating pack item


andrew-bierman (changes requested):
It looks like packs are being copied, but they do not show up in the user's packs. Testing on pretty much every pack in feed

andrew-bierman:
We might need to revise our logic in here. How about this as we discussed:

What if we kept owner_id as the single source of truth for who owns a pack. But then we utilized owners array for actual co-owners of a pack (like google docs style, other owners can edit the actual pack). Then when someone goes to copy a pack, we actually create a brand new instance with them as owners and the same items array. We could just add a forked_from property to track the origin pack, to show stats.

andrew-bierman (approved):
lgtm, nice job

PR #141, opened 2023-07-17
save the photon res in db

ab958:
hi andrew,

i have done the photon part, now data is being saved in db, if you approved this i will move towards osm part.
thanks

### DIFF
--- a/server/controllers/getOsmController.js
+++ b/server/controllers/getOsmController.js
@@ -3,7 +3,7 @@ import axios from "axios";
 import Way from "../models/osm/wayModel.js";
 import Node from "../models/osm/nodeModel.js";
 import mongoose from "mongoose";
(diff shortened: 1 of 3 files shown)


andrew-bierman:
This looks good to me so far, no critiques from my end

andrew-bierman:
One thing I thought of, is that we should take a look at mongosm and how it creates the mongo schema for OSM data, while our solution will be a little more custom (due to geojson conversion) - this could provide a good base to check if our schema has the right info from OSM side.

andrew-bierman (changes requested):
@ab958 any reason I should merge this before you do the Overpass logic ? I think we should probably just wait until that is done and working

ab958:
i have added function to check nodes, if present in database, then don't create a new node instead just used id.

andrew-bierman:
Error during bulk insert, still getting CastErrors frequently 

Screenshot 2023-07-22 at 4 32 03 PM

ab958:
can you please more detail about that issue. want to recreate this error as i am unable to recreate it.

PR #143, opened 2023-07-17
Destination

ArbazYousuf:
1. Pass the correct lat lng in Map
2. check some dispatch of destination


andrew-bierman (changes requested):
See comments

PR #152, opened 2023-07-20
Mikibo gpx download upload

mikib0:
- move the gpx to geojson conversion from server to client 
- added a button for gpx upload (both platforms)
- downloading a gpx now uses the trail name as the file name


andrew-bierman (changes requested):
This is awesome. Let's just resolve merge conflict and then we can merge this into andrew_testing

andrew-bierman (approved):
nice work

PR #155, opened 2023-07-23
Pack copy templates with selective items as well as all items

aezazali1997:
added the logic to slect items that needs to be copied


aezazali1997:
> added the logic to slect items that needs to be copied
The user can select the pack which he want to copy. user taps on the copy icon. The details pack screen is shown. The user can then select the items which he want to be included in pack. then after selecting. he can press copy to copy the pack to his own profile

PR #156, opened 2023-07-23
DropdownIssue Solved

zainaming:


andrew-bierman:
Opening pack from pack dropdown now throws this 
"Log 1 of 30

Uncaught Error
Objects are not valid as a React child (found: object with keys {_id, name, owner_id, is_public, favorited_by, favorites_count, createdAt, owners, grades, scores, type, items, total_weight}). If you meant to render a collection of children, use an array instead."

andrew-bierman (changes requested):
See comment, i'm getting that error on both web and iOS for pack selection. Some others work well though. 

To recreate go to /trips page and try to select a pack

andrew-bierman:
This still throws virtualized list error on pages :(

client/components/Dropdown.js
```diff
@@ -1,19 +1,60 @@
...
+  ItemStyle,
+  ItemTextStyle,
+  SelectTextStyle,
+  SelectStyle,
   placeholder,
   width,
   style,
 }) => {
-  // console.log("data in dropdown", data);
+  console.log("data in dropdown", style);
```
andrew-bierman, line 22: if i `return null` here then the virtualized list error goes away FYI. So something here is still causing our virtualized list issue

andrew-bierman (changes requested):

andrew-bierman (approved):
nice work

PR #159, opened 2023-07-23
Mikibo refactor profile

mikib0:
- unified the UserDataCard component to work on both platforms;
-  fixed the pack details page to work -it previously had a bug that causes the whole app to crash
- fixed an error that get thrown on mobile, from the table 
- fixed some styling issues on mobile in the ScoreContainer component


andrew-bierman (approved):
lgtm

PR #166, opened 2023-07-25
user SignIn and Register issue solved

zainaming:


andrew-bierman (commented):
Just curious, what is the purpose of this change? It seems like we are just moving our password hashing logic to a new place? Or am I missing something

WW43l:
we not only changed these place i add const salt = await bcrypt.genSalt(parseInt(JWT_SECRET)); thats line of code in our previous code we missing this thats way its gave error for some users whoes password in bcrypt form

andrew-bierman (approved):

PR #170, opened 2023-07-26
Feature vl

umerbaigh:


andrew-bierman:
So aside from all the formatting, it seems you added 2 flatlists, are there any other notable changes you made?

umerbaigh:
Hi @andrew-bierman ,
I successfully resolved the error mentioned in the assigned ticket, which was "VirtualizedLists should never be nested inside plain ScrollViews with the same orientation - use another VirtualizedList-backed container instead." The issue occurred due to the presence of nested scroll views, where a scroll view was placed inside another scroll view. To fix this problem, I made the following changes:
1. Trips Screen:
   - Replaced the previous usage of a FlatList with a ScrollView, which resolved the error.
2. Packs Screen:
   - Replaced the previous implementation of select items using a ScrollView with a custom Select component, which eliminated the error.
As a result of these modifications, the app is now functioning smoothly without encountering these errors. If you have any further questions or concerns, feel free to reach out. Thank you for your cooperation!

PR #173, opened 2023-07-27
Integrate adapter into redux store

abdallahalkashawy:
all adapters are working properly except DestinationStore as it's not full functional and you can show all commit messages


andrew-bierman (approved):
lgtm

PR #178, opened 2023-07-29
Fix/misc fixes across fe be

andrew-bierman:


andrew-bierman:
@vipin4771234 see changes to userModel and userController

PR #184, opened 2023-07-30
Feature native

vipin4771234:


client/hooks/useGetUser.js
andrew-bierman, line 1: We should probably be doing this from redux store

client/components/user/UserProfileContainerMobile.js
andrew-bierman, line 1: Can we just merge this into the UserProfileContainer instead of having 2 components? Will be easier to maintain

andrew-bierman (changes requested):
Small changes requested, otherwise this is good to go!

client/hooks/useGetUser.js
vipin4771234: Actually i have done it using redux store. useGetUser.js is not used.

vipin4771234 (commented):
I have given feedback

client/components/user/UserProfileContainerMobile.js
vipin4771234: ### I have merged it into single

andrew-bierman:
Can we do this with the ProfileContainer? I feel like we are rewriting a lot of code between ProfileContainer and UserProfileContainer

andrew-bierman (approved):

PR #191, opened 2023-07-31
user setting task done

AhsanDev404:
-> frontend and page of user setting
-> responsive 
-> web only


AhsanDev404:
Okay I will check and again send you PR.

PR #196, opened 2023-08-01
code updated

owaisanwar:


owaisanwar:
on destination page, i am using the condition of geometry type also to check if it is point or not and on trip page we will have the line string.
const isDestinationMap = (selectedSearchResult, type) => {
  return selectedSearchResult?.geometry?.type === 'Point' && type === DESTINATION
}

client/components/map/NativeMap.native.js
```diff
@@ -171,7 +174,7 @@ function NativeMap({ shape: shapeProp }) {
       ></View>
     );
   }
-
+  const pointLatLong = selectedSearchResult?.geometry?.coordinates
```
andrew-bierman, line 177: Can't we just get this from the shape (geojson) ?

client/components/map/WebMap.web.js
```diff
@@ -165,7 +173,10 @@ const WebMap = ({ shape: shapeProp }) => {
   }, [mapFullscreen]);
 
   useEffect(() => {
-    if (map.current) {
+    if(map.current && isDestinationMap(selectedSearchResult, type)) {
+      addPoints(map.current);
+    }
+    else if (map.current && selectedSearchResult?.geometry?.type !== 'Point') {
```
andrew-bierman, line 179: Also should be in geojson (shape)

client/components/map/WebMap.web.js
```diff
@@ -229,6 +240,14 @@ const WebMap = ({ shape: shapeProp }) => {
     });
   };
 
+
+  const addPoints = (mapInstance) => {
```
andrew-bierman, line 244: same as above

andrew-bierman (changes requested):
Overall looks nice, i think we just need to refactor to use the shape instead of these new props.

PR #199, opened 2023-08-01
details page copy button and view owner button change

vipin4771234:


andrew-bierman (changes requested):
cards in feed still show View Owner

andrew-bierman (changes requested):
this shows the pack name not the username of owner

andrew-bierman:
The cards in the /packs and profile should say something than view owner if the user is the owner

vipin4771234:
> cards in feed still show View Owner

We don't have owner name access on feed data array. I have put the link only in pack/trip details page only. fetching owner for each card on feed page would be unnecessary.

vipin4771234 (commented):
Changed View owner to view `${username}` and if user of pack is same as view user then show 'Your Profile'

andrew-bierman:
> 

We could populate user on mongo query in backend ? Then we would have access. Just an idea.

vipin4771234 (commented):
done

vipin4771234:
Andrew Please verify new merge

PR #200, opened 2023-08-01
Saving trip and view it

abdallahalkashawy:


andrew-bierman (changes requested):
Are maps displaying the right locations on your end ? For me it's just showing the default shape for everything

andrew-bierman:
> Are maps displaying the right locations on your end ? For me it's just showing the default shape for everything

Example in the create trip page, and viewing a saved trip. It is always the default shape ?

abdallahalkashawy:
It wasn't working from begining , I didn’t change anything for that logic

PR #208, opened 2023-08-03
add static map image preview

mikib0:
image


andrew-bierman (approved):
nice, still need to debug why certain map types are not working. but you did good on this. looks great

PR #210, opened 2023-08-03
Umang

gucci1909:
middleware , redux token in api's and Basic Item Recommendation Algorithm


client/store/destinationStore.js
```diff
@@ -6,7 +6,11 @@ export const processGeoJSON = createAsyncThunk(
   "destination/processGeoJSON",
   async (data, { rejectWithValue }) => {
     try {
-      const response = await axios.post(`${api}/osm/process/geojson`, data);
+      const token = await AsyncStorage.getItem('userToken');
```
andrew-bierman, line 9: We should just add a middleware to our store like this, would cut down on a top of repeated code. Rough idea:

`const authMiddleware = (storeAPI) => (next) => (action) => {
  if (action.type.includes('pending')) {
    const token = storeAPI.getState().auth.token;

    if (token) {
      // Attach token to headers
      action.meta.arg.headers = { ...action.meta.arg.headers, Authorization: `Bearer ${token}` };
    }
  }

  return next(action);
};
`

andrew-bierman (changes requested):
See comments

andrew-bierman:
I'm getting 'invalid token' error for all API requests. Are you experiencing same thing ?

gucci1909:
> I'm getting 'invalid token' error for all API requests. Are you experiencing same thing ?

yes @andrew-bierman same

andrew-bierman:
These should be split into different pull requests

PR #214, opened 2023-08-04
Restful api

gucci1909:
Restful-api


andrew-bierman (changes requested):
Code not functional on preview deploy.

See previous comments on redux logic

andrew-bierman:
Quick reference on RESTful best practices:

Resource Naming: Use nouns, not verbs, to name your resources. For example, /getUsers or /deleteUser is incorrect, whereas /users is the correct way to name the resource. HTTP methods should define the operation being performed on the resource.

    Use HTTP Methods Appropriately:
        GET: Retrieve a specific resource (by id) or a collection of resources.
        POST: Create a new resource.
        PUT: Update a specific resource (by id) entirely.
        PATCH: Update a part of the resource.
        DELETE: Remove a specific resource by an id.

    Use HTTP Status Codes Correctly:
        2xx Success (e.g., 200 OK, 201 Created, 204 No Content)
        3xx Redirection (e.g., 301 Moved Permanently, 304 Not Modified)
        4xx Client errors (e.g., 400 Bad Request, 401 Unauthorized, 404 Not Found)
        5xx Server errors (e.g., 500 Internal Server Error, 502 Bad Gateway)

    URL Structure: Ideally, a URL should look like this: https://api.example.com/v1/users/1234. Note the following aspects:
        Use HTTPS whenever possible.
        API versioning (v1) is important for maintaining backwards compatibility.
        User-friendly URL segments: URLs should be human readable and as friendly as possible.
        Resource IDs: /users/1234 is the correct way of accessing a specific user.

 Controller Structure: Controllers should be lean, having the sole responsibility of delegating requests to the appropriate service functions. They receive a request, extract necessary data from it, call the relevant services, and then send a response.
 
 
 We are not using ids in routes properly, we often just do it via the body, but that does not follow best practices.

PR #215, opened 2023-08-04
Umang

gucci1909:
middlewares in backend , redux api calls by token and restricted by role and basic Item Role Algorithm - 3 tickets in one PR because I can't raise different PR for these tickets as these three are connected to one another


andrew-bierman (changes requested):
You have not addressed any of the comments I made on the original PR. This should be done via a redux middleware, not doing token logic on every thunk. 

Use redux persist instead also.

PR #218, opened 2023-08-04
Itinerary

gucci1909:
itinerary ticket done


client/store/authStore.js
```diff
@@ -23,6 +24,9 @@ export const signUp = createAsyncThunk(
         email,
         password,
       });
+      const token = response.data.user.token;
+      await AsyncStorage.setItem('userToken', token); // Save token in local storage
```
andrew-bierman, line 28: Redux persist as we discussed ?

client/store/chatStore.js
```diff
@@ -14,7 +15,11 @@ export const getUserChats = createAsyncThunk(
   "chat/getUserChats",
   async (userId) => {
     try {
-      const response = await axios.get(`${api}/openai/user-chats/${userId}`);
+      const token = await AsyncStorage.getItem('userToken');
```
andrew-bierman, line 18: middleware instead or adding this to every thunk

client/components/trip/itineraryTrip.js
```diff
@@ -0,0 +1,91 @@
+import React, { useState, useEffect } from "react";
```
andrew-bierman, line 1: does not seem to be updating live across 2 instances on my end, no way to add itinerary to an existing trip

server/models/itineraryModel.js
```diff
@@ -0,0 +1,20 @@
+import { Schema, model } from "mongoose";
```
andrew-bierman, line 1: seems like we are missing info to correlate to an existing trip

andrew-bierman (changes requested):
See comments. This feature is not actually usable in it's current form. Looks like you added all the thunk stuff we discussed changes for also to this PR

PR #219, opened 2023-08-04
tool tip added

gucci1909:


andrew-bierman (changes requested):
We need a reusable tooltip component ?

gucci1909:
> We need a reusable tooltip component ?

@andrew-bierman done

andrew-bierman:
When you do the reusable components ticket, we should convert this to tamagui instead of native base as that is what we are transitioning to.

PR #223, opened 2023-08-05
itinerary added

gucci1909:
itinerary added

### DIFF
--- a/client/components/trip/createTrip.js
+++ b/client/components/trip/createTrip.js
@@ -16,15 +16,15 @@ import { Platform, StyleSheet } from "react-native";
 
 import { useEffect, useState } from "react";
 
-import { useDispatch, useSelector } from "react-redux";
+import { useSelector } from "react-redux";
 
 import { GearList } from "../GearList";
 
 import { SaveTripContainer } from "./createTripModal";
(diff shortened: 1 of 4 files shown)


andrew-bierman:
See previous comments please https://github.com/andrew-bierman/PackRat/pull/218

client/components/trip/createTrip.js
```diff
@@ -55,17 +48,6 @@ export default function Trips() {
     setParksData(parksObject);
   }, [parksObject]);
 
-  useEffect(() => {
```
andrew-bierman, line 58: Why remove ?

server/models/itineraryModel.js
```diff
@@ -0,0 +1,20 @@
+import { Schema, model } from "mongoose";
```
andrew-bierman, line 1: How are we connecting itineraries to their corresponding trip ?

andrew-bierman (changes requested):

andrew-bierman:
Itineraries belong to a specific trip. They are not a requirement of a trip, but an additional option. I think for simplicity, a user could add them after a trip is created, on the edit phase, rather than during creation.

PR #224, opened 2023-08-05
index-mongodb added

gucci1909:
index-mongodb added

### DIFF
--- a/server/models/itemModel.js
+++ b/server/models/itemModel.js
@@ -24,5 +24,8 @@ const ItemSchema = new Schema(
   { timestamps: true }
 );
 
+ItemSchema.index({name:1});
+
+
 const Item = myDB.model("Item", ItemSchema);
 export default Item;
--- a/server/models/osm/nodeModel.js
+++ b/server/models/osm/nodeModel.js
@@ -14,6 +14,9 @@ const NodeSchema = Schema(
   { timestamps: true }
 );
 
+// Add an index on the "id" field
+NodeSchema.index({ id: 1 });
+
 // add a to JSON method to the schema
 NodeSchema.method("toJSON", async function () {
   const { _id, ...object } = this.toObject();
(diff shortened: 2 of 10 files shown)


server/models/tripModel.js
```diff
@@ -34,6 +34,18 @@ const TripSchema = new Schema(
   { timestamps: true }
 );
 
+// Create an index on the "owner_id" field
```
andrew-bierman, line 37: We could do this from within Schema definition I think

server/models/itemModel.js
```diff
@@ -24,5 +24,8 @@ const ItemSchema = new Schema(
   { timestamps: true }
 );
 
+ItemSchema.index({name:1});
```
andrew-bierman, line 27: This should probably be a compound index, with text

server/models/osm/nodeModel.js
```diff
@@ -14,6 +14,9 @@ const NodeSchema = Schema(
   { timestamps: true }
 );
 
+// Add an index on the "id" field
+NodeSchema.index({ id: 1 });
```
andrew-bierman, line 18: is a 2d index possible here ?

server/models/osm/relationModel.js
```diff
@@ -22,6 +22,9 @@ const RelationSchema = new Schema(
   { timestamps: true }
 );
 
+// Add an index on the "osm_id" field
```
andrew-bierman, line 25: is a 2d index possible here ?

server/models/osm/wayModel.js
```diff
@@ -23,6 +23,10 @@ const WaySchema = new Schema(
   { timestamps: true }
 );
 
+// Add an index on the "osm_id" field
```
andrew-bierman, line 26: is a 2d index possible here ?

andrew-bierman:
There are opportunities to use the different type of mongo indexes in here. Also maybe some wildcard indexes for frequently searched things

andrew-bierman (changes requested):

PR #225, opened 2023-08-05
3 ticket - middleware backend, redux header token , basic item algo

gucci1909:
3 tickets - 1 . middleware in backend , 2 - redux headers token , 3 - basic algo 
sorry right now I think we can work without redux persistor , in future I will make the changes that you are saying


andrew-bierman:
See previous comments please https://github.com/andrew-bierman/PackRat/pull/210

andrew-bierman:
I think this should be a redux middleware from our store, rather than manually repeating same token Bearer Log logic in every thunk call.

andrew-bierman (changes requested):
Change to redux middleware for token, rather than manually  repeating Bearer logic for every thunk.

andrew-bierman:
Rough draft of something like this for our store.js file:

````
const authMiddleware = (storeAPI) => (next) => (action) => {
if (action.type.includes('pending')) {
const token = storeAPI.getState().auth.token;

if (token) {
  // Attach token to headers
  action.meta.arg.headers = { ...action.meta.arg.headers, Authorization: `Bearer ${token}` };
}

}

return next(action);
};

PR #228, opened 2023-08-07
Search items on pack

aezazali1997:
in progress. kindly check wether i am going to right direction or should i make changes


client/components/item/searchItem.js
```diff
@@ -0,0 +1,204 @@
+import {
```
andrew-bierman, line 1: This looks good. i think we should make SearchInput.js reusable and utilize it here also

andrew-bierman (approved):
looks good. note to self that we should extract search input into reusable component.

PR #229, opened 2023-08-08
Zain branch

zainaming:


client/screens/Navigation.js
```diff
@@ -27,6 +27,7 @@ import Drawer from "./Drawer";
 import { Link, useRouter, usePathname } from "expo-router";
 import { hexToRGBA } from "../utils/colorFunctions";
 import UseTheme from "../hooks/useTheme";
+import { useRoute } from "@react-navigation/native";
```
andrew-bierman, line 30: We should probably be using expo router here

andrew-bierman: I think?

zainaming: yes

zainaming: it's only import neither use it anywhere.

andrew-bierman:
Do you get this on your end ?
Screenshot 2023-08-09 at 7 18 19 PM

andrew-bierman:
> Do you get this on your end ? Screenshot 2023-08-09 at 7 18 19 PM

I'm still getting this just FYI. It works better though

PR #231, opened 2023-08-09
Mikibo save geojson for destination

mikib0:
Save geojson to db directly instead of modelling osm types. This will simplify lots of things and fix lots of the issues faced with saving osm while also preventing duplication.

### DIFF
--- a/server/controllers/tripController.js
+++ b/server/controllers/tripController.js
@@ -1,7 +1,5 @@
 import Trip from "../models/tripModel.js";
-import Node from "../models/osm/nodeModel.js";
-import Way from "../models/osm/wayModel.js";
-import Relation from "../models/osm/relationModel.js";
+import GeoJSON from "../models/geojsonModel.js";
 
 export const getPublicTrips = async (req, res) => {
   try {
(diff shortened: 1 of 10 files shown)


andrew-bierman:
Thanks for this, but looks like we are missing the geojson model you referenced. Did you intend to have a model for this, or just include geojson  in one of the other schemas?

server/controllers/tripController.js
```diff
@@ -1,7 +1,5 @@
 import Trip from "../models/tripModel.js";
-import Node from "../models/osm/nodeModel.js";
-import Way from "../models/osm/wayModel.js";
-import Relation from "../models/osm/relationModel.js";
+import GeoJSON from "../models/geojsonModel.js";
```
andrew-bierman, line 2: missing file

andrew-bierman (changes requested):

server/controllers/tripController.js
mikib0: forgot to include it in the commit 😅

andrew-bierman (approved):
nicely done

PR #239, opened 2023-08-12
Server folder restructure

raghavyuva:
[TODO TASK](https://github.com/users/andrew-bierman/projects/3/views/1?pane=issue&itemId=34973687)

- File structure changed.
- Controller has folders 
     - functions as files
     - index.js file which exports all the controller functions 
 - Each functions under controller now has a JSDoc string which makes it more documented.
 - Index.js file under server now has less cluttered codes
 - Created a helpers folder which now exports helping functions to index.js file
 - PR is divided into smaller chunks of commits so that review can be hastle free


andrew-bierman (approved):
awesome job, this is much cleaner

PR #241, opened 2023-08-14
fix: redux slice fixes

NoumanSakhawat:
[Github Project Ticket](https://github.com/users/andrew-bierman/projects/3/views/1?filterQuery=react&pane=issue&itemId=35524369)

single pack redux issue fix

### DIFF
--- a/client/store/singlePackStore.js
+++ b/client/store/singlePackStore.js
@@ -10,7 +10,7 @@ import axios from "axios";
 
 import { api } from "../constants/api";
 
-const singlePackAdapter = createEntityAdapter({
+export const singlePackAdapter = createEntityAdapter({
   selectId: (singlePack) => singlePack._id,
 });
 
@@ -47,6 +47,19 @@ export const selectItemsGlobal = createAsyncThunk(
   }
 );
 
+export const updateExistingSinglePack = createAsyncThunk(
+  "packs/updateExistingSinglePack",
+  async (pack) => {
+    return pack;
+  }
+);
+export const addNewSinglePack = createAsyncThunk(
+  "packs/addNewSinglePack",
+  async (pack) => {
+    return pack;
+  }
+);
+
 const singlePackSlice = createSlice({
   name: "singlePack",
   initialState: initialState,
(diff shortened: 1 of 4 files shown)


andrew-bierman:
We should be doing this in the packs store, that is where the state is not updating. Go to the create trip component (/trips) and try to add a pack and edit an item and monitor redux dev tools

andrew-bierman (changes requested):
Please see packs store not single packs store

andrew-bierman:
Note, 'packs' slice is for the user's packs (where they would perform CRUD). 'singlePack' is for viewing details of a pack, whether or not the user is the owner. I think that is the confusion.

NoumanSakhawat:
> > Note, 'packs' slice is for the user's packs (where they would perform CRUD). 'singlePack' is for viewing details of a pack, whether or not the user is the owner. I think that is the confusion.
> 
> @andrew-bierman but I guess we are showing the details from the `singlePack` so we don't need to handle `pack` as whenever user came from pack, he has the updated value.

@andrew-bierman and can you please run the app and test it using my branch `fix/redux-slice-fixes` am pretty much it will work as expected

client/store/singlePackStore.js
```diff
@@ -79,6 +92,34 @@ const singlePackSlice = createSlice({
       .addCase(selectItemsGlobal.rejected, (state, action) => {
         state.isLoading = false;
         state.error = action.error.message;
+      })
+      .addCase(updateExistingSinglePack.fulfilled, (state, action) => {
+        const index = state.singlePack.items.findIndex(
+          (x) => x._id === action.payload._id
+        );
+        if (index === -1) return;
+        const data = JSON.parse(JSON.stringify(state.singlePack));
```
andrew-bierman, line 101: This feels a bit hacky doing the parse and stringify, why ?

client/store/singlePackStore.js
```diff
@@ -10,7 +10,7 @@ import axios from "axios";
```
andrew-bierman, line 10: Ok i get what you are saying and i understand your code here is accomplishing the fix, but should we just move this whole slice into packs and add a new state for singlePack ? 

The main issue i have is having 2 thunks that call the same api route, and i feel like we should consolidate. Currently, the packs in packs slice is user packs, but singlePack could contain another user's pack. 

I feel like we should not have to update the entire pack to just update one pack item.

andrew-bierman (changes requested):

andrew-bierman:
/
```javascript
      .addCase(editPackItem.fulfilled, (state, action) => {
        const newItem = action.payload;
        const packIds = newItem.packs; // packIds is an array of pack Ids

        packIds.forEach((packId) => {
          console.log("packid", packId);
          // loop through each packId
          const existingPack = state.entities[packId];
          console.log("existingPack", existingPack);
          if (!existingPack) {
            return;
          }

          const updatedItems = existingPack.items.map((item) =>
            item._id === newItem._id ? newItem : item
          );
          console.log("updatediTEMS", updatedItems);

          packsAdapter.updateOne(state, {
            id: packId,
            changes: { items: updatedItems },
          });
        });

        state.isLoading = false;
        state.error = null;
        state.isOpenEditModal = false;
        state.update = !state.update;
      })
``` 

This is the logic not properly updating.

PR #242, opened 2023-08-14
Dev shams

shamsKhokhar56:
Hi Andrew, I have fixed the carousal component for the feed section in dashboard. Also fixed the unexpected increase in height for the IOS when the card is pressed.


andrew-bierman (changes requested):
This is way cleaner, but on web in mobile view, buttons disappear. Also get a inque key warning in console on iOS emulate for the carousel component.

Let's get that fixed then this is good to merge

shamsKhokhar56:
I have fixed that warning issue on ios and had to change the icons because expo web was not able to render that. Couldn't find the fix of that so i changed the icons. Kindly review.

andrew-bierman (approved):

PR #248, opened 2023-08-15
Migration of server files to typescript

raghavyuva:
[TYPESCRIPT SUPPORT FOR SERVER](https://github.com/users/andrew-bierman/projects/3/views/1?filterQuery=types&pane=issue&itemId=31662802)
- All files are now under src folder.
- tsconfig.json assumes src folder as root directory
- Nodemon.json is created so that when you run nodemon all the scripts inside nodemon.json gets executed
- All files do support typescripts
- Strict typechecking is set to false so that migration can be done rigorously
- It's assumed that the developers will continue to add types and follow typescript syntaxes going forward in the repo


andrew-bierman (approved):
nicely done

PR #250, opened 2023-08-16
Service files to handle model manipulations from the controllers

raghavyuva:
- Service is a folder which contains different subfolder according to controller
- Each folder under services has its own .service.ts file which exports all the other functions related to that folder
- Controllers are updated to make use of the service files that are created


andrew-bierman:
Thank you, one think i feel like could be a helpful change would be utilizing classes for the services. 

Example:

Base Model class with all basic mongoose CRUD and query logic so we can utilize it in all classes.

Service file(s) for each model, extending base and implementing specific logic.

All DB interactions should now go through a service class.

andrew-bierman:
> Thank you, one think i feel like could be a helpful change would be utilizing classes for the services.
> 
> Example:
> 
> Base Model class with all basic mongoose CRUD and query logic so we can utilize it in all classes.
> 
> Service file(s) for each model, extending base and implementing specific logic.
> 
> All DB interactions should now go through a service class.

Open to discussion on this approach though.

andrew-bierman (approved):
nicely done

PR #251, opened 2023-08-16
added global confirmation box with custom props

aezazali1997:
added global confirmation component


andrew-bierman:
Just wondering, why not use our CustomModal instead of another dependency ?

aezazali1997:
> Just wondering, why not use our CustomModal instead of another dependency ?

right. i didnt think of that while coding. let me refactor it again

PR #271, opened 2023-08-19
Class Based Architecture in Server 

raghavyuva:
[FEATURE](https://github.com/users/andrew-bierman/projects/3?pane=issue&itemId=36285833)
- Controllers index.ts file is now a class which extends Base class of the controller
- src/routes/index.ts is now registers all the route controllers in the application
- server.ts now has all the code related to server
- App itself is a class which will be used in the server.ts file.
- App class has middlewares routes and errorHandler functions
- routes files are moved to respective controller classes


raghavyuva:
closing this pr now, since we do not plan to make this conversion

PR #275, opened 2023-08-19
2 tickets of middleware

gucci1909:
2 tickets of middleware - frontend and backend both


andrew-bierman (changes requested):
I'm still getting a lot of invalid token errors for some reason

PR #283, opened 2023-08-22
direction button added to map container

owaisanwar:


andrew-bierman:
i'm getting taken to bad addresses on web at least, haven't tried mobile. Example search 'old rag' on web and then navigate

PR #286, opened 2023-08-22
Mikibo user settings page

mikib0:
goto /settings


andrew-bierman:
First name last name not prepopulating FYI. Can we connect these to real requests via auth store for updating name/email. Don't worry about storing profile picture until we get S3 set up

andrew-bierman:
If the routes for changing name and email look to be too much work, that is fine we can merge this in for UI if needed

mikib0:
> First name last name not prepopulating FYI. 

Turns out they don't exist on the user model, instead it's just a name field. I think it would be better to have firstName & lastName instead of just name 🤔.

> Can we connect these to real requests via auth store for updating name/email. Don't worry about storing profile picture until we get S3 set up

Yes, I planned to take the whole updating functionality (including password change) as another ticket. Will create another branch for that.

Let me refine the UI a bit and add section for password change, then we can merge it.

andrew-bierman:
Great thanks. Username would be good to have also

mikib0:
> Great thanks. Username would be good to have also

Yep, also added in the new commit.

mikib0:
image

andrew-bierman:
> > First name last name not prepopulating FYI. 
> 
> 
> 
> Turns out they don't exist on the user model, instead it's just a name field. I think it would be better to have firstName & lastName instead of just name 🤔.
> 
> 
> 
> > Can we connect these to real requests via auth store for updating name/email. Don't worry about storing profile picture until we get S3 set up
> 
> 
> 
> Yes, I planned to take the whole updating functionality (including password change) as another ticket. Will create another branch for that.
> 
> 
> 
> Let me refine the UI a bit and add section for password change, then we can merge it.

Agreed let's change to first and last and migrate existing if needed.

andrew-bierman:
> image
> 
> 

This is looking good 👍

PR #287, opened 2023-08-22
Redux persist auth encryption Async storage using AES encryption

raghavyuva:


andrew-bierman:
Nice work. This is good, just wanted to flag a couple small things for discussion:

- Should we be using user specific keys in addition to our ENCRYPTION KEY ?
- Implementing salting/IV to increase security. Brainstorming below 

```javascript
  transforms: [
    {
      in: (state, key) => {
        const userKey = getUserEncryptionKey();  // Retrieve the user-specific key
        const iv = CryptoJS.lib.WordArray.random(128 / 8);  // Generate a random IV

        const encryptedState = CryptoJS.AES.encrypt(JSON.stringify(state), userKey, { iv: iv }).toString();
        
        // Store IV with the encrypted state, separated by a delimiter for later extraction
        return iv.toString(CryptoJS.enc.Hex) + ":" + encryptedState;
      },
      out: (encryptedStateWithIv) => {
        const userKey = getUserEncryptionKey();  // Retrieve the user-specific key

        const parts = encryptedStateWithIv.split(":");
        const iv = CryptoJS.enc.Hex.parse(parts[0]);
        const encryptedState = parts[1];

        const bytes = CryptoJS.AES.decrypt(encryptedState, userKey, { iv: iv });
        const decryptedState = JSON.parse(bytes.toString(CryptoJS.enc.Utf8));
        
        return decryptedState;
      }
    }
  ]
```

client/package.json
```diff
@@ -49,7 +49,9 @@
     "@tanstack/react-query": "^4.27.0",
     "@tmcw/togeojson": "^5.7.0",
     "axios": "^1.3.5",
+    "babel-plugin-module-resolver": "^5.0.0",
```
andrew-bierman, line 52: not sure we need this library?

client/store/store.js
```diff
@@ -48,11 +49,56 @@ const rootReducer = combineReducers({
   userStore,
 });
 
-// configure persist store and whitelist reducers
+function getUserEncryptionKey() {
+  return process.env.ENCRYPTION_KEY ?? "fd2aafcbf5c604e2613b68e9ef794dc5a18a2715927e9a968d3054c1f2be345a"
```
andrew-bierman, line 53: should we just generate instead of hardcoding?

andrew-bierman:
Commenting to check if all is good here or if you need more feedback from me

client/package.json
raghavyuva: removed that out, but on my local it gave error that this module was not installed

client/store/store.js
raghavyuva: the problem in generating is you can generate the key while encrypting because the state of the user is plain text 
but when you want to decrypt you need user information but the user information itself is encrypted then how you'll get the key?

PR #294, opened 2023-08-23
Create a Multi Page 'Sliding' Component

shamsKhokhar56:
Hi Andrew, Kindly review this.


andrew-bierman:
Thank you for working on this, but i think we should be utilizing the MultiStepForm for these features, as I don't want to over complicate our carousel component.

PR #295, opened 2023-08-24
FIX - reactivity in packdetails component

raghavyuva:
- Updated EditableInput to take updates whenever global state updates
- Updated PackDetails components useffect dependency array ( removed updated constant  which was a hack around this issue)
- Updated AuthSlice with spread operator of the state and new state while returning ( which solves issue `A case reducer on a non-draftable value must not return undefined`
   ```
   .addCase(signUp.rejected, (state, action) => {
        return { ...state, loading: false, error: action.payload };
      })
    ```
- Fixed SinglePack fetching issue 
```.addCase(selectItemsGlobal.fulfilled, (state, action) => {
        const updatedSinglePack = {
          ...state.singlePack,
          items: [...state.singlePack.items, action.payload.data],
        };
        state.singlePack = updatedSinglePack;

        state.isLoading = false;
        state.error = null;
      })
```

### DIFF
--- a/client/components/EditableText/index.js
+++ b/client/components/EditableText/index.js
@@ -1,4 +1,4 @@
-import React, { useState } from "react";
+import React, { useState,useEffect } from "react";
 import { TextInput } from "react-native";
 import { theme } from "../../theme";
 import { useDispatch } from "react-redux";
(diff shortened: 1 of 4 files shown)


andrew-bierman:
I'm not getting reactivity in /trips pack section or on pack details page. Are you having same isssue on your end ?

raghavyuva:
> I'm not getting reactivity in /trips pack section or on pack details page. Are you having same isssue on your end ?

for me it is working fine

PR #309, opened 2023-08-26
add icon for navigating to settings from profile

mikib0:
image


mikib0:
Looks like this has already been added.

PR #318, opened 2023-08-27
Multi Step Form

shamsKhokhar56:
I have implemented steps for create a trip. I haven't comment the progress bar that was previously added. Review and tell me if you need anything to change.

### DIFF
--- a/client/components/SearchInput.js
+++ b/client/components/SearchInput.js
@@ -116,13 +116,13 @@ export const SearchInput = ({ onSelect, placeholder }) => {
     return () => clearTimeout(timeout);
   }, [selectedSearch, selectedSearchResult, dispatch]);
 
-/**
- * Handles the click event on a search result.
- *
- * @param {object} result - The search result object.
- * @param {number} index - The index of the search result.
- * @return {undefined} This function does not return anything.
- */
+  /**
+   * Handles the click event on a search result.
+   *
+   * @param {object} result - The search result object.
+   * @param {number} index - The index of the search result.
+   * @return {undefined} This function does not return anything.
+   */
   const handleSearchResultClick = (result, index) => {
     const {
       properties: { name, osm_id },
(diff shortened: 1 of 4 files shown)


andrew-bierman:
Ok this is much cleaner. The only thing we are missing is an optional summary that we can display on the side with large screens. My vision is that this have data similar to what we show in the save modal, and just provide a high level overview of what was selected. This does not need to be too complex, but I think it will help the UX.

andrew-bierman (changes requested):
Add an additional optional sidebar for summary data, most of this work has been done already. Otherwise looks good to go

shamsKhokhar56:
Yes, there was a side component that i have commented. If it works, We can also show the selected data in the last step, i.e., Save Trip. As there is only a save trip button and it looks weird.

andrew-bierman:
Perfect thank you. I think that is only thing that is needed, if quick fix, please add to this PR. Then I will merge in. 

If not quick, no worries about it. 

Appreciate your work on this.

andrew-bierman:
> Yes, there was a sidebar component that i have commented. If it works, We can also show the selected data in the last step, i.e., Save Trip. As there is only a save trip button and it looks weird.

Yes that would be great addition. That save button actually opens a modal with the selected data - but with new UI we should just show that data in that step.

shamsKhokhar56:
Okay. Will do it tomorrow in my time. I will create a PR for that.

shamsKhokhar56:
One last thing. Do we still need to have a sidebar for large screens ?

andrew-bierman:
> One last thing. Do we still need to have a sidebar for large screens ?

I think it would be helpful to have that yes. Does not need to be complex - I think if you just add back in the current sidebar and touch up styling that will be sufficient.

shamsKhokhar56:
Andrew, please review for the changes.

andrew-bierman (approved):

PR #319, opened 2023-08-27
trip issue fixed

aezazali1997:
fix trips not loading


andrew-bierman (approved):
nice work, added some changes and looks like it is working on my end :)

PR #321, opened 2023-08-27
basic offline support ticket with some reduced repeating code

owaisanwar:


client/components/item/AddItemGlobal.js
```diff
@@ -39,18 +39,27 @@ export const AddItemGlobal = ({
...
-    dispatch(
-      addItemsGlobal({
-        name,
-        weight,
-        quantity,
-        type: categoryType,
-        unit,
-      }),
-    );
+    console.log(isConnected, 'is connected');
```
andrew-bierman, line 42: Ok this is great, but I imagine we will be repeating it a ton throughout app. As every async dispatch is going to need this check. 

Should we make a reusable component to check for isConnected, then do dispatch logic accordingly?

andrew-bierman (approved):

andrew-bierman:
Nice work. See comment for future discussion

PR #335, opened 2023-08-29
WIP - Zod validation integrated by replacing celebrate library

raghavyuva:
- Replaced celebrate library by zod library
- Converted all the validations to make use of zod validations


andrew-bierman:
Nice. We should probably move object Id and parser logic out of each of these files so we can reuse them. Otherwise looks solid.

Any luck with trpc ?

andrew-bierman (approved):

andrew-bierman:
Merging for convenience, but can add your new changes as well

PR #337, opened 2023-08-29
adding a global item does not update the table in the items screen

witehound:


witehound:
@andrew-bierman this task can be merged!

PR #339, opened 2023-08-29
Update validations for login / sign up. Add proper checks for constraints we enforce

witehound:


server/src/middleware/validators/userRoutesValidator.ts
```diff
@@ -18,6 +18,7 @@ export const userSignUp = celebrate({
     name: Joi.string().required(),
     email: Joi.string().required(),
     password: Joi.string().required(),
```
andrew-bierman, line 20: We have constraints we should be validation here also, such as a couple password things

andrew-bierman (changes requested):

server/src/middleware/validators/userRoutesValidator.ts
witehound: Like what?

andrew-bierman: See the userModel we have a few requirements

andrew-bierman: @witehound this is still pending

witehound: Those are all the required fields! And now the function works as it should really!

witehound: @andrew-bierman just checked and I think this issue has been properly resolved

andrew-bierman: Are there any changes in this PR that are not already in andrew_testing ? After syncing branch it does not show any touched files

witehound: Oh Yh when we had the zod validation issue the last branch that was working was this branch so I reverted to this branch and started building till you finally updated andrew_testing? So most of the lastest brabches I have already have this same changed !

andrew-bierman:
Closing because nothing seems to be changing anymore. Note to self, add user validations similar to model ie password

PR #340, opened 2023-08-29
Nearby Trips / Nearby Parks Should Be Carousel Cards Instead of Dropdowns

witehound:
Fix broken carousel ui
Screenshot 2023-08-29 at 23 06 34


andrew-bierman:
This ticket was to improve the UI so the card looks better. Or at least similar to some of our better cards we use in app like on feed. We might need to pass in more data which we will have from overpass query

PR #346, opened 2023-08-30
Usestyles hook

witehound:
created a useStyle hooks that helps to reduce use of style(), also reuseing styles around the app increases UI consistency. review if i should proceed or comment for improvements.


client/hooks/useStyles.js
```diff
@@ -0,0 +1,337 @@
+import useTheme from './useTheme';
```
andrew-bierman, line 1: This needs to be more dynamic. I was thinking we can pass in object with styles and get back stylesheet version. Needs to account for currenttheme. Otherwise this file is just going to become even larger

andrew-bierman (changes requested):

client/hooks/useStyles.js
andrew-bierman: Open to other solutions though

witehound: Yh because of the size I stopped to actually ask the question yes using an object would be a better idea but for now won’t be able to merge or get rid of similar styles.

andrew-bierman:
This is what I was going for. Nice work. Just need to run it on my local, but from reading through code looks good

andrew-bierman:
I'm getting a couple errors on imports. Can you click around FE when you get a chance and check all the touched components styles, then I can merge in

witehound:
@andrew-bierman definetely could not do that before due to the zod validator I’d fine a way

andrew-bierman (approved):
nice

PR #349, opened 2023-09-01
Search Filters for Searching Our mongo

witehound:
- updated server service 
- upddated fe code


andrew-bierman:
Nice work here, the only thing i think Most Items is not functioning for me? Are you seeing this work on your side ?

witehound:
Yes! I have packs with zero items and then one with an actual item and it filters are you using the demo ?

witehound:
@andrew-bierman fixed looks like already existing code had a bug

andrew-bierman (approved):
looks good

PR #355, opened 2023-09-03
Ratpack UI testing in trips page

witehound:
This was very hectic work! please leave a review


andrew-bierman:
The name is PackRat not RatPack

andrew-bierman:
Font looks strange on my side ? It somehow went to a times new roman style font

andrew-bierman:
The actual packrat ui components looks like you just re-exported tamagui for each one?

andrew-bierman:
Search input is not fitting properly and looks worse than how we had it previously

andrew-bierman:
Dropdown component could be good ?

witehound:
@andrew-bierman yes i was expecting comments like this, so i had to import the ratpack ui like that because sometimes the object itself is needed instead of just a component i was looking to talk this over with you! so sorry about the naming issue, so unlike native base tamagui does not support things like placing an object in input so i had to be creative and use just css! we just have to decide on how to proceed from here

andrew-bierman:
So you would just re-export all the tamagui components we would use, and then pretty much redo UI using tamagui is that correct ? just trying to understand your approach. I assume for things like cards we might want to add some of our own props to simplify this

witehound:
@andrew-bierman yes! That’s what I’ve done for now it’s written this way for constant improvement as agreed!

andrew-bierman:
- [ ] Fix naming issue
- [ ] Search input seems strange compared to original UI/UX
- [ ] Dropdown?

witehound:
@andrew-bierman fixed all

PR #358, opened 2023-09-04
Add Scoring Logic to Trips, Similar to Pack Scoring task one

witehound:
Needs logic on actual method to use to score packs


andrew-bierman (changes requested):
Missing the scoring logic ?

andrew-bierman:
The scoring logic here should take into account trip weather, hike location, pack weight and score.  Just need a draft for now similar to packs but a few more variables to account for

witehound:
@andrew-bierman currently in the trips page I don’t think it’s the weather for days of the trip that gets sent back to the db! I’m trying to confirm so I’ll be able to fix! Not sure how we’d go about rating hike location would it be same score as long as there’s an hike location?

andrew-bierman:
> @andrew-bierman currently in the trips page I don’t think it’s the weather for days of the trip that gets sent back to the db! I’m trying to confirm so I’ll be able to fix! Not sure how we’d go about rating hike location would it be same score as long as there’s an hike location?

Yes that is a good point. 

For the time being, let's just assume the weather associated with a trip is accurate (I know it is not for trips planned further than 5 days out)  and take it into account. Just need some boilerplate logic for scoring and then this is good to merge in. We can come improve the logic later on.

andrew-bierman:
Need to add frontend score button like we have for packs

andrew-bierman:
Added boilerplate scoring logic for testing, let's just get the button and score to show up on FE and then this is good to merge in.

client/components/ScoreContainer.js
```diff
@@ -161,7 +163,7 @@ export default function ScoreContainer({ type, data, isOwner }) {
     useTheme();
   const styles = useCustomStyles(loadStyles);
   const id = data._id;
-  const totalScore = data.totalScore;
+  const totalScore = computeTotalScore(data?.scores);
```
andrew-bierman, line 166: This is going to break the pack scoring, let's use the virtual property you set up instead?

client/utils/scores.js
```diff
@@ -0,0 +1,8 @@
+export const computeTotalScore = (scores) => {
```
andrew-bierman, line 1: Is the virtual property not already doing this ?

witehound: Yes couldn’t get it on the trips so I just replaced it

PR #361, opened 2023-09-05
Users cannot signin or signup with google

witehound:


andrew-bierman:
Google auth is working fine for me on main branch, both with existing and new accounts, are you having issues?

witehound:
@andrew-bierman it wasn’t working on Andrew_testing because of some fields that are required and also if you logged in successfully you can’t make any request to the  backend! But I fixed all of those !

PR #363, opened 2023-09-07
Store User Uploaded Maps in DB

witehound:
- [ ] user upload fe
- [ ] user upload be


andrew-bierman:
The flow of adding a gpx map and saving it is not working for me. I see the map saves something but its not saving/displaying the correct uploaded geojson.

andrew-bierman:
Go to create trip page and try to create a trip with a custom map and it does not seem to work on my side ?

andrew-bierman:
Steps I was taking:
Create Trip page -> upload a valid geojson, map appears blank. Or try to search for a place then upload a map, the map seems to disappear ?

PR #367, opened 2023-09-10
Rough draft of shared packrat packages folder

andrew-bierman:


andrew-bierman:
We need to get this working, then just re-export it from the current server validators for the time being.

server/src/controllers/favorite/addToFavorite.ts
```diff
@@ -3,7 +3,7 @@ import { addToFavoriteService } from '../../services/favorite/favorite.service';
 import { UserNotFoundError } from '../../helpers/errors';
 import { responseHandler } from '../../helpers/responseHandler';
 import { publicProcedure } from '../../trpc';
-import * as validator from '../../middleware/validators/index';
+import * as validator from '../../../../packages/src/validations';
```
andrew-bierman, line 6: We should be using the @packrat/packages as a dependency within the server and client, not just doing ../../ to get to the packages directory

server/tsconfig.json
```diff
@@ -8,7 +8,6 @@
     ],
     "allowJs": true,
     "outDir": "dist",
-    "rootDir": "src",
```
andrew-bierman, line 11: src should still be rootDir

PR #384, opened 2023-09-19
Feat/generalized-functions-rebase

andrew-bierman:


andrew-bierman (commented):
Sign in page keeps popping in front of all pages.

Local storage is not defined, should we use hook ? What if its in secure storage ?

client/hooks/useStorageState.ts
```diff
@@ -39,18 +30,17 @@ export function useStorageState(key: string): UseStateHook<string> {
...
-        }
-      } catch (e) {
-        console.error('Local storage is unavailable:', e);
-      }
-    } else {
-      SecureStore.getItemAsync(key).then(value => {
-        setState(value);
-      });
+
+    const getUser = async () => {
```
andrew-bierman, line 34: We should just store the token not the entire user object

client/trpc.ts
```diff
@@ -1,15 +1,22 @@
 import { createTRPCProxyClient, httpBatchLink } from '@trpc/client';
 import type { AppRouter } from 'server/src/routes/trpcRouter';
+import AsyncStorage from '@react-native-async-storage/async-storage';
 import { api } from './constants/api';
 
 console.log('api', api);
 
+const getUser = async (key) => {
```
andrew-bierman: Token instead of user i think, but let me know if there's a reason to store whole user

client/hooks/useStorageState.ts
```diff
@@ -39,18 +30,17 @@ export function useStorageState(key: string): UseStateHook<string> {
...
-    if (Platform.OS === 'web') {
-      try {
-        if (typeof localStorage !== 'undefined') {
-          setState(localStorage.getItem(key));
-        }
-      } catch (e) {
-        console.error('Local storage is unavailable:', e);
-      }
-    } else {
-      SecureStore.getItemAsync(key).then(value => {
```
andrew-bierman: Why remove secure store for native? I think we should bring this part back, leave expo AsyncStorage for web

client/hooks/useStorageState.ts
```diff
@@ -9,27 +8,19 @@ function useAsyncState<T>(
...
-      } else {
-        localStorage.setItem(key, value);
-      }
-    } catch (e) {
-      console.error('Local storage is unavailable:', e);
-    }
-  } else {
+  try {
     if (value == null) {
-      await SecureStore.deleteItemAsync(key);
```
andrew-bierman: We should bring this part back and use SecureStore on native, and keep what you added with AsyncStorage for web

andrew-bierman (commented):
Minor changes for just storing token instead of user obj, and bringing back the securestorage from expo on native side. Otherwise looks good

client/hooks/useStorageState.ts
witehound: Okay that works I thought because we already used it like that then that should work just fine

PR #385, opened 2023-09-19
loader fixed with styles

aezazali1997:
### Added loader for table and text for individual Packs

image


client/components/pack_table/Table.tsx
```diff
@@ -339,7 +340,7 @@ export const TableContainer = ({
   }
   console.log(heading);
 
-  if (isLoading) return <Loading />;
+  if (!data?.length) return <TableLoader />;
```
andrew-bierman, line 343: A pack with no items will always show loader

client/components/skeleton/index.tsx
```diff
@@ -0,0 +1,24 @@
+import { Center, Skeleton, VStack } from 'native-base';
```
andrew-bierman, line 1: Move to ui folder, try to minimize native base usage. We might want to pull the native base skeleton source code into our own ui component

aezazali1997: i tried a few methods but tamagui does not comes with a built-in Skeleton. so i moved the native-base code into packrat.

andrew-bierman (approved):

PR #386, opened 2023-09-19
WIP - Zod frontend Validations

raghavyuva:


andrew-bierman:
I'm getting a lot of errors importing, so unable to test. How are you building packages on your side ? Is it different than the build command?

andrew-bierman (changes requested):
We could be using the reusable form to validate all of these in real time, not just on submission (which we can also do). 

But the form should allow for us to show errors for the user based on zod validation.

PR #389, opened 2023-09-21
fixed trip bugs

witehound:


andrew-bierman:
Looks like create trip throws a 401 when trying to create

andrew-bierman:
Nearby Parks not populating

client/store/trailsStore_copy.js
```diff
@@ -19,16 +18,17 @@ export const fetchTrails = createAsyncThunk(
     if (lon) {
       params += `&lon=${lon}`;
     }
-    const radius = 500;
```
andrew-bierman, line 22: Why no radius ?

witehound: I was getting more trails result without the radius than with the radius do you think it’s okay if Trails result don’t always come in for every location?

PR #392, opened 2023-09-26
Feat/experimenting with trpc query

andrew-bierman:
All things from redux store that need to be moved over 

- [ ]   auth: authReducer, // IGNORE FOR NOW
- [ ]   dropdown: dropdownReducer, // DOES NOT NEED TO BE MIGRATED
- [ ]   search: searchReducer, // DOES NOT NEED TO BE MIGRATED
- [ ]   weather: weatherReducer,
- [x]   parks: parksReducer,
- [x]   items: itemsReducer,
- [x]   packs: packsReducer,
- [x]   trips: tripsReducer,
- [x]   favorites: favoritesReducer,
- [x]   singlePack: singlePackReducer,
- [x]   singleTrip: singleTripReducer,
- [x]   feed: feedReducer,
- [x]   gpx: gpxReducer,
- [x]   destination: destinationReducer,
- [x]   chat: chatReducer,
- [x]   globalItems,
- [x]   userStore,
- [ ]   offlineQueue, // HANDLE DIRECT WITH REACT QUERY 
- [ ]   progress: progressReducer, // ignore
- [x]   trails : trailsReducer

Reference this for create, edit, delete actions - https://www.whitespectre.com/ideas/how-to-build-offline-first-react-native-apps-with-react-query-and-typescript/


andrew-bierman:
looks like i did not push up changed from last night, but i was having some success doing:
```typescript
utils.getPackById.invalidate()
```

andrew-bierman:
For trips/destination - there is a bit more work required here. we need to dumbify our search input so that it is just UI and accepts props with what to do. This will no longer use a global store for thing that probably should be the responsibility of parent component. 

Then we can add custom trpc hooks per usual

andrew-bierman:
Adding items to a pack via the /trips screen is broken. It works from the standard pack details screen. Also packs are appearing empty in the /trips screen and not showing their items.

andrew-bierman:
https://github.com/andrew-bierman/PackRat/issues/512#issue-2049860566

andrew-bierman:
We need to figure out our state management solution for something like create trip. How should we pass around the data for this? I'm open to pass around state props if we can avoid drilling too deep. 

We have some stuff in redux still (ie weather state, search state (photon)), but i think we could probably optimize this. Maybe we move state logic and mutations to more defined redux slices (ie set weather object into the createTrip redux store. So all async stuff would go through trpc query but for some of these more complex forms we use redux? But I think we could also probably accomplish a lot of this with just useState also. 

Or custom hook with useState could solve too honestly.

PR #399, opened 2023-09-28
updated branch with hooks of pack screens

owaisanwar:


client/app/_layout.js
```diff
@@ -13,20 +14,45 @@ import { SessionProvider } from '../context/auth';
 import { ThemeProvider } from '../context/theme';
 import FlashMessage from 'react-native-flash-message';
 import Footer from '../components/footer/Footer';
+import { queryTrpc, getToken } from '../trpc';
+import { httpBatchLink } from '@trpc/client';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { api } from '~/constants/api';
 
 export default function HomeLayout() {
+  const queryClient = new QueryClient()
```
andrew-bierman, line 23: we should move this to to its own TrpcQueryProvider context just for cleaner code

client/hooks/packs.js
```diff
@@ -0,0 +1,72 @@
...
+            
+        default:
+            return { data: null, erro : null };
+    }
+    return data   
+}
+
+export const useAddNewPack = () => {
+    // console.log(obj);
+    const mutation = queryTrpc.addPack.useMutation();
```
andrew-bierman, line 70: is this standard naming practice for react query custom hooks? genuine question i'm not sure

client/hooks/packs.js
```diff
@@ -0,0 +1,72 @@
...
+import { queryTrpc } from "../trpc";
+import { tr } from "date-fns/locale";
+
+export const usePacks =  (queryString, ownerId, feedType, selectedTypes) => {
+    console.log(feedType,'feed type', ownerId,queryString, 'query string',selectedTypes)
+    let  data = [];
+    let isLoading = true;
+    switch (feedType) {
+        case 'public':
+            const publicPacks = queryTrpc.getPublicPacks.useQuery({ queryBy:queryString ? queryString:  'Favorite' }, {
```
andrew-bierman, line 11: We should at the very least be destructuring isLoading and error from the react query responses. To make it more like the redux store

client/hooks/packs.js
andrew-bierman: We should at the very least be destructuring isLoading and error from the react query responses. To make it more like the redux store

client/components/pack/AddPack.js
```diff
@@ -28,7 +28,7 @@ export const AddPack = () => {
   const error = useSelector((state) => state.packs.error);
 
   const isError = error !== null;
-
+  const { mutation } = useAddNewPack()
```
andrew-bierman, line 31: naming convention

andrew-bierman (commented):
overall good just clarifying a few things

andrew-bierman:
Uncaught Error
publicTrips.map is not a function

client/hooks/weather.js
```diff
@@ -0,0 +1,29 @@
...
+
+export const useFetchWeatherWeak = (latLng) => {
+    const { lat, lon } = latLng;
+    const isEnabled = Boolean(lat && lon)
+    const { refetch, data, error, isLoading, isError  } = queryTrpc.getWeatherWeek.useQuery({ lat, lon }, {
+        enabled : isEnabled,
+        refetchOnWindowFocus: false,
+        keepPreviousData: true,
+    });
+    const isOwner = data && user && data.owner_id === user._id;
```
andrew-bierman, line 25: user is undefined here, maybe we should use our matches current user hook for this ?

PR #403, opened 2023-09-28
TRPC error handling in controllers

raghavyuva:
- Try catch blocks
- Throw TrpcErrors 
- Error handling in comparision with REST API's


andrew-bierman:
for the error throwing, is the idea if it makes it to one of our controls and we still error, then that truly is an internal service error 500? as validations and such should stop it beforehand ?

PR #407, opened 2023-09-29
Feed rebuilt

witehound:


client/screens/feed/Feed.js
```diff
@@ -1,19 +1,17 @@
 import React, { useEffect, useState } from 'react';
 import { useDispatch, useSelector } from 'react-redux';
+import { Box, Text, IconButton } from 'native-base';
```
andrew-bierman, line 3: these should be coming from packrat ui

andrew-bierman:
> Left out the drop down component as for some reason tamag ui might have actually made the whole page slower please let me know

i think our current dropdown needs to be swapped since we use that quite a bit

client/screens/feed/Feed.js
witehound: will fix

witehound:
> > Left out the drop down component as for some reason tamag ui might have actually made the whole page slower please let me know
> 
> i think our current dropdown needs to be swapped since we use that quite a bit

so theres one in the user-validation we might be able to build on

andrew-bierman:
> > > Left out the drop down component as for some reason tamag ui might have actually made the whole page slower please let me know
> > 
> > 
> > i think our current dropdown needs to be swapped since we use that quite a bit
> 
> so theres one in the user-validation we might be able to build on

yes i agree, we could move that one to packrat-ui and add in relevant logic from the current components/dropdown file

witehound:
> > > > Left out the drop down component as for some reason tamag ui might have actually made the whole page slower please let me know
> > > 
> > > 
> > > i think our current dropdown needs to be swapped since we use that quite a bit
> > 
> > 
> > so theres one in the user-validation we might be able to build on
> 
> yes i agree, we could move that one to packrat-ui and add in relevant logic from the current components/dropdown file

i dont quite get what you mean?

andrew-bierman:
packrat-ui should have a select / dropdown component. that is what we should use on this feed component. but we need to be mindful about the existing /components/dropdown as it is widely used - but we are transitioning it to be a packrat-ui component now

andrew-bierman:
there is a memory leak somewhere in this new component, keeps getting slow and slowly crashing for me

andrew-bierman:
something in the search filter header section possibly ?

witehound:
I thought it was the drop down component when I added the tamag ui select component so I took it out, left a message that it should be a seperate component. I’d try different things and see what works. I think tamag ui might just be slow

witehound:
removed the tamagui switch it was really buggy i assume that is where the problem originated from

andrew-bierman:
Note - re-render possibly in feed card ?

PR #418, opened 2023-10-04
Feat 34410273

sk04852:


server/src/models/geojsonModel.ts
```diff
@@ -39,6 +39,9 @@ const GeoJSONSchema = new Schema(
   },
 );
 
+// Adding an index to the 'id' field
+GeoJSONSchema.index({ id: 1 }, { unique: true });
```
andrew-bierman, line 43: shouldn't this be a geospatial index ?

server/src/models/itemModel.ts
```diff
@@ -23,5 +23,7 @@ const ItemSchema = new Schema(
   { timestamps: true },
 );
 
+ItemSchema.index({ location: '2dsphere' });
+
```
andrew-bierman, line 27: compound index with text ?

PR #421, opened 2023-10-07
feat:implement opyimistic update

salman-aziz-4425:


client/hooks/useGetPacks.js
```diff
@@ -1,14 +1,25 @@
...
-// export default function useGetPacks(owner_id) {
-//   const { data, isError, error, isLoading } = useQuery({
-//     queryKey: ["packs"],
-//     queryFn: async () => await fetcher(`${api}/pack/${owner_id}`),
-//     enabled: Boolean(owner_id),
-//   });
+export default function useGetPacks({ ownerId, queryString }) {
+  console.log( ownerId)
+  const { data, isError, error, isLoading } = useQuery(
+    ["packs", ownerId], // Use ownerId and queryString as part of the queryKey
```
andrew-bierman, line 7: Any reason to not use the trpc query here ?

client/hooks/useAddPack.js
```diff
@@ -1,28 +1,59 @@
...
+      onMutate:async (newPack) => {
+         await queryClient.cancelQueries(["packs", newPack.owner_id])
+        // Step 1: Define optimistic update
+        const optimisticUpdate = {
+          ...newPack,
+          id: Date.now(),
+        };
+        console.log("Optimistic update")
+        console.log( newPack)
+        const oldQueryData= queryClient.getQueryData(["packs", newPack.owner_id]);
```
andrew-bierman, line 30: Do we have to be using the actual queryClient here ? As opposed to the trpc instance you were using in others

client/hooks/useGetPacks.js
andrew-bierman: And isn't this the same logic as what we have in the /hooks/packs directory ?

andrew-bierman:
Is there anything remaining on this PR that needs to be included or can we close ?

andrew-bierman:
> Is there anything remaining on this PR that needs to be included or can we close ?

@salman-aziz-4425 see comment

PR #423, opened 2023-10-08
Optimistic updates

salman-aziz-4425:


client/components/pack/AddPack.js
```diff
@@ -39,7 +39,7 @@ export const AddPack = () => {
    * @return {void}
    */
   const handleAddPack = () => {
-    dispatch(addPack({ name, owner_id: user?._id, is_public: isPublic }));
+    mutation.mutate({ name, owner_id: user?._id, is_public: isPublic })
```
andrew-bierman, line 42: we should be returning mutation.mutate in the hook so we can call addPack like you did for edit pack item

client/hooks/packs/useAddNewPack.js
```diff
@@ -1,6 +1,61 @@
+import { getQueryKey } from '@trpc/react-query';
+import { queryClient } from '~/constants/queryClient';
```
andrew-bierman, line 2: if we need query client we should be importing the trpc one for consistency (not your fault i know its confusing how we have multiple created in app at the moment)

client/components/pack/AddPack.js
salman-aziz-4425: Ok i will push it again along with delete pack item

client/hooks/packs/useAddNewPack.js
salman-aziz-4425: I mistakenly added this i am only using use context majorly will remove it

PR #430, opened 2023-10-18
trpc updates

salman-aziz-4425:


andrew-bierman:
Ok so trips seems to definitely be relying on store, which we are no longer updating for some of these,  that is why its not saving

client/components/user/UserDataCard.js
andrew-bierman, line 1: what is going on in here ?

client/screens/feed/Feed.js
```diff
@@ -15,13 +15,7 @@ import {
   Flex,
 } from 'native-base';
 import { AntDesign } from '@expo/vector-icons';
-import {
-  StyleSheet,
-  FlatList,
-  View,
-  ScrollView,
-  RefreshControl,
```
andrew-bierman, line 23: Why removing this ?

client/screens/feed/Feed.js
```diff
@@ -174,18 +169,18 @@ const Feed = ({ feedType = 'public' }) => {
...
   });
   const [searchQuery, setSearchQuery] = useState('');
-
+  const [Data,setData]=useState(data)
   const [refreshing, setRefreshing] = useState(false);
 
   const dispatch = useDispatch();
   const ownerId = useSelector((state) => state.auth.user?._id);
-  // const publicPacksData = useSelector((state) => state.feed.publicPacks);
-  // const userPacksData = useSelector(selectAllPacks);
```
andrew-bierman, line 183: We don't want to use redux for these things

client/components/user/UserDataCard.js
salman-aziz-4425: basically favorited_count was empty but we dont need this as fevorited_by have ids of all users that like that pack so we can show count just by checking length of favorited_by

client/screens/feed/Feed.js
salman-aziz-4425: didnt remove this just due to identation changes its showing its their

client/screens/feed/Feed.js
salman-aziz-4425: yeah most of them is not being used now but for removing auth we have to carefully remove it

client/components/user/UserDataCard.js
andrew-bierman: But it looks like we are deleting the UserDataCard ? Is this intentional ?

salman-aziz-4425: No i was just checking the functionality of pages in that scenario i removed then added back

salman-aziz-4425: No we aren't deleting anything just pack card favorites count wasn't updating so i used favorited_by array length for count

client/components/trip/createTripModal.js
```diff
@@ -68,18 +69,24 @@ const NumberInput = (props) => {
   );
 };
 
-export const SaveTripContainer = ({ dateRange }) => {
+export const SaveTripContainer =({ dateRange }) => {
   const [isSaveModalOpen, setIsSaveModalOpen] = useState(false);
   const weatherObject = useSelector((state) => state.weather.weatherObject);
-  const search = useSelector((state) => state.search.selectedSearchResult);
```
andrew-bierman, line 74: we need to keep the selected search bar result somewhere else i think. do we have a hook for this search already ?

salman-aziz-4425: Yeah we already have but weather information was stored in state weather so that's why state.search wasn't have information regarding geoJson and we were getting errors

salman-aziz-4425: previously we were accessing osm_id and other information from state.search but we were setting state.weather.selectedSearch while selecting destination

client/components/trip/createTripModal.js
```diff
@@ -115,18 +122,18 @@ export const SaveTripContainer = ({ dateRange }) => {
...
-
+    
     const data = {
       name,
       description,
       start_date: startDate,
       end_date: endDate,
       destination: search.properties.name,
-      geoJSON,
+      geoJSON:JSON.stringify(geoJSONObject),
```
andrew-bierman, line 136: why are we doing this stringify / parse logic ?

client/components/trip/createTripModal.js
andrew-bierman: yea we need to move that logic out of the redux store though i think ? its a bit confusing currently with both in play

salman-aziz-4425: Yeah in this scenario for example we can create a small custom hook for fetching weather information instead we dispatch it first and then use it

PR #431, opened 2023-10-20
WIP - redid expo prebuild. Then cleared all caches, including using react native cli and metro directly. made expo/head platform specifc.

andrew-bierman:


andrew-bierman:
I’m trying to make our builds more reliable and not break as often with different features. I have fixed them on this branch (working on 2 systems for me, Mac and Linux)

We need reliable GitHub actions to ensure the app can be build multiple ways. eas, Xcode, android / gradle. This is close just need to get over finish line

PR #444, opened 2023-10-25
Andrew_testing

nandwere:
Bottom Tabs Update


andrew-bierman:
Ok this looks nice but i'm worried we are duplicating a lot of files? Many of the files in (tabs) are also in (app). Is there a cleaner way to handle ?

nandwere:
You're correct. There is option of restructuring and having reusable components. Working on that.

PR #448, opened 2023-10-28
Trpc hook

HaykHunanyan:


client/components/SearchInput.js
```diff
@@ -26,10 +26,9 @@ import { fetchWeather, fetchWeatherWeek } from '../store/weatherStore';
 import useCustomStyles from '~/hooks/useCustomStyles';
 
 export const SearchInput = ({ onSelect, placeholder }) => {
-  const [searchString, setSearchString] = useState('');
+  const [searchString, setSearchString] = useState({ name: '' });
```
andrew-bierman, line 29: Why are we making it an object instead of string here?

client/Podfile.example
andrew-bierman, line 1: we don't want to remove these example files

client/hooks/useApiCall.js
```diff
@@ -0,0 +1,27 @@
+import { useState } from 'react';
```
andrew-bierman, line 1: I don't understand the purpose of this hook. We should be using the trpc wrapper for react-query, which will handle all of these states.

client/hooks/useDestination.js
```diff
@@ -0,0 +1,35 @@
+import { trpc } from '../trpc';
```
andrew-bierman, line 1: The purpose of this task is to use the trpc react query wrapper, why are we not using that here ? See the other hooks we have done on the trpc query branch for reference

client/store/weatherStore.js
```diff
@@ -16,6 +16,8 @@ const weatherAdapter = createEntityAdapter();
 const initialState = weatherAdapter.getInitialState({
```
andrew-bierman, line 16: We are trying to move away from redux for these types of requests. So why are we adding things into here ?

andrew-bierman (changes requested):
See comments

client/components/SearchInput.js
HaykHunanyan: Because in there you used the (set timeout) javascript function and it's not correct to do that, it's bad code and I remove that setTimeout and used react hooks(useState) with object for tracking searching name .
Using an object as state provides the flexibility to handle multiple values within a single state variable. This structure can be beneficial when dealing with complex components where different fields or pieces of data need to be managed collectively.

client/Podfile.example
HaykHunanyan: I used Readme to run a project

PR #450, opened 2023-10-29
Bottom/drawer navigation

nandwere:
- Added Mobile Navigation drawer.
- bottom tabs refactor
- Migrated yarn to berry.
- Other build fixes.


andrew-bierman:
Is there a reason why we should not just switch our web navigation from stack to tab? I'm concerned about how much duplication we have in our routes with this

client/components/dashboard/QuickActionSection.js
```diff
@@ -27,10 +28,11 @@ const QuickActionsSection = () => {
    * @param {string} action - The selected action.
    */
   const handleActionSelect = (action) => {
+    const isWeb = Platform.OS == 'web';
     if (action === 'createPack') {
-      router.push('/pack/create');
+      router.push(isWeb ? '/pack/create' : '/tabs/pack/create');
```
andrew-bierman, line 33: i'm a bit worried about logic like this. we should decide on one way to go here. i think we can go all in on tabs

andrew-bierman: but we don't want the route to be /tabs/* in the URL

PR #457, opened 2023-11-01
gpx integration with trpc

alexander01202:


andrew-bierman (changes requested):
We are trying to move away from redux for server state and actions. We should put this into a react query style hook, using trpc wrapper. See other ones done on the trpc query branch.

alexander01202 (commented):
Is there any other route that should have it's cache invalidated?

andrew-bierman:
Ok this is close but missing the hook. 

We should be using the trpc react query router. 

See other hooks we have already done in /hooks/subdirectories 

So we should have useGpx file or something in /hooks that we call.

andrew-bierman:
> Is there any other route that should have it's cache invalidated?

not for this gpx slice, but some of the others yes

andrew-bierman (approved):

andrew-bierman:
looks like we currently handle client side which is fine and probably better. just found this 

```javascript
      <MapButtonsOverlay
        mapFullscreen={mapFullscreen}
        enableFullScreen={enableFullScreen}
        disableFullScreen={disableFullScreen}
        handleChangeMapStyle={handleChangeMapStyle}
        fetchLocation={fetchLocation}
        styles={styles}
        downloadable={downloadable}
        downloading={downloading}
        navigateToMaps={openMaps}
        onDownload={fetchGpxDownload}
        handleGpxUpload={async () => {
          console.log('clikedd');
          try {
            const result = await DocumentPicker.getDocumentAsync({
              type: 'application/gpx+xml',
            });
            console.log('result', result);
            if (result.type === 'success') {
              const base64Gpx = result.uri.split(',')[1];
              const gpxString = atob(base64Gpx);
              const parsedGpx = new DOMParser().parseFromString(gpxString);
              const geojson = toGeoJSON(parsedGpx);
              setShape(geojson);
            }
          } catch (err) {
            Alert.alert('An error occured');
          }
        }}
        shape={shape}
      />
```

PR #458, opened 2023-11-02
Feat/prisma proof of concept

andrew-bierman:
Auto gen pulls in a ton of collections we don't need to migrate from dev db.

Collections to keep:

- [ ] openai
- [ ] geojson
- [ ] item category
- [ ] item
- [ ] pack
- [ ] template
- [ ] trip
- [ ] user


awais-codes:
I can remove the models that are not necessary. Also the schema.prisma file may need formatting using prisma format command

https://www.prisma.io/docs/reference/api-reference/command-reference

I can add prisma CLI scripts to package.json as well. Your call @andrew-bierman

andrew-bierman:
Those are all good ideas. 

Before we get too far along with integration I want to test that we can replace a few of our routes with prisma ORM and see that everything works fine in wrangler. 

But everything you said above is good.

awais-codes:
> Those are all good ideas. 
> 
> Before we get too far along with integration I want to test that we can replace a few of our routes with prisma ORM and see that everything works fine in wrangler. 
> 
> But everything you said above is good. 

Sure. Hope it works fine

awais-codes:
> Those are all good ideas. 
> 
> Before we get too far along with integration I want to test that we can replace a few of our routes with prisma ORM and see that everything works fine in wrangler. 
> 
> But everything you said above is good. 

Thanks. 

Hope it goes well. I think this doc would be helpful to you in this case specially step#6 and step#7

https://www.prisma.io/docs/guides/deployment/edge/deploy-to-cloudflare-workers

PR #462, opened 2023-11-03
feat: add prisma schema

awais-codes:


server/src/prisma/schema.prisma
```diff
@@ -0,0 +1,252 @@
...
+  @@map("templates")
+}
+
+model Trip {
+  id          String   @id @default(auto()) @map("_id") @db.ObjectId
+  v           Int      @map("__v")
+  name        String
+  // should be JSON?
+  description String
+  // should be JSON?
```
andrew-bierman, line 126: Just curious - why do you think these should be JSON?

server/src/prisma/schema.prisma
```diff
@@ -0,0 +1,252 @@
...
+}
+
+//  ███    ███  ██████  ██████  ███████ ██      ███████
+//  ████  ████ ██    ██ ██   ██ ██      ██      ██
+//  ██ ████ ██ ██    ██ ██   ██ █████   ██      ███████
+//  ██  ██  ██ ██    ██ ██   ██ ██      ██           ██
+//  ██      ██  ██████  ██████  ███████ ███████ ███████
+
+/// TODO: replicate mongoose static methods
+/// TODO: check prisma capability to regex validate type and id_ fields
```
andrew-bierman, line 17: we can handle this regex thing with zod if we need to, don't worry too much about this one

awais-codes: Oh alright

server/src/prisma/schema.prisma
awais-codes: I saw these fields had stringified JSON values in the collection for the provided connection string. So I was just curious why not store JSON instead?

PR #465, opened 2023-11-04
swipe screens

nandwere:
- Added Get Started Swipe Screens.


client/components/dashboard/QuickActionSection.js
```diff
@@ -27,10 +28,11 @@ const QuickActionsSection = () => {
    * @param {string} action - The selected action.
    */
   const handleActionSelect = (action) => {
+    const isWeb = Platform.OS == 'web';
     if (action === 'createPack') {
-      router.push('/pack/create');
+      router.push(isWeb ? '/pack/create' : '/tabs/pack/create');
```
andrew-bierman, line 33: I thought we had removed the need for conditional routing?

nandwere: Yes. We removed that. Seems whatever you have is no updated

PR #466, opened 2023-11-04
initial next config

andrew-bierman:
needs more integration from t4 stack, ability to display our react native screens

- T4 Stack
- Tamagui starter 
- Solito starter

good references


andrew-bierman:
Notes - This is from solito starter. It works but not with our tamagui components. We need to merge in config from tamagui starter and follow docs to set up in project. Additionally, the t4 stack is another create reference for how they set up this config.

PR #472, opened 2023-11-06
tripStore trpc conversion 

alexander01202:


client/hooks/feed/index.js
```diff
@@ -1,6 +1,6 @@
 import { usePublicFeed } from './publicFeed';
 import { useUserPacks } from './../packs';
-import { useUserTrips } from './../trips';
```
andrew-bierman, line 3: Why this change ?

client/components/EditableText/index.js
```diff
@@ -57,8 +58,8 @@ export const EditableInput = ({
             setEditTitle(false);
             titleRef.current.style =
               'font-size:20px !important;font-weight:bold;color: #22c67c;';
-
-            dispatch(editTrip(tripDetails));
+            useEditTrips(tripDetails)
```
andrew-bierman, line 61: pretty sure we should be destructuring the mutation func out of here and calling it ? or am I mistaken

client/hooks/feed/index.js
alexander01202: The trpc query hook was for singleTripsStore but was saved in a folder called trips instead of singleTrips, which was stopping me from saving my own trpc hook (which was the actually trpc trips hook) in a trips folder. I don't know if my explanation makes sense to you?

client/components/EditableText/index.js
alexander01202: My bad. I'd fix that right away

alexander01202:
Fixed now andrew.

client/hooks/feed/index.js
andrew-bierman: i see what you are saying, but user trips will return more than 1 trip, so i think we should name it something else ? or maybe i'm misunderstanding

alexander01202: No it wouldn't. All I did was rename the folder, so I can use the previous folder name for mine instead. There are no double files or double calls of `useUserTrips`. It's still the exact same folder/file structure, just with a different folder name. (from trips -> singletrips)

andrew-bierman: Got it thanks for clarifying

PR #483, opened 2023-11-11
WIP - Server, Mongoose -> Prisma. Hono integration. 

andrew-bierman:


awais-codes:
1. Setup standalone tRPC setup i.e no elysia.
2. Replaced bcrypt with bcryptjs
3. updated server package.json script and .ignore
4. commented user routes
5. updated prisma imports to use edge based client.

andrew-bierman:
Below are some queries that I'm running into issues with on my side, I think mostly due to response size/length. We should be more explicit about the data we need (based on frontend). List is in progress still. 

We have strict [limitations](https://www.prisma.io/docs/accelerate/limitations) using the accelerate/data proxy from prisma.

- [ ] getPacks
- [ ] getPublicPacks
- [ ] getTrips
- [ ] getPublicTrips
- [ ] getUserFavorites

andrew-bierman:
- [ ] getPhotonResults cannot read properties of undefined

andrew-bierman:
Note, need to add values from env to env.dev.vars in wrangler toml for development environment. Be sure not to commit these to repo.

andrew-bierman:
DATABASE_URL will need to be updated to a prisma connection pooling URL for something like prisma accelerate to work properly on edge

server/src/prisma/statics/node.ts
```diff
@@ -0,0 +1,29 @@
+const findOrCreateMany = async (ids, nodes) => {
```
yasTheDreamer, line 1: to fix error : `this `might be undefined
```suggestion
async function findOrCreateMany(ids, nodes) {
```

yasTheDreamer (changes requested):

PR #487, opened 2023-11-11
Refactor to use our tamagui wrapper instead of native-base

mbonete:
Related to task:
https://github.com/users/andrew-bierman/projects/3/views/1?filterQuery=&pane=issue&itemId=34309618


client/screens/feed/Feed.tsx
```diff
@@ -74,72 +69,70 @@ const FeedSearchFilter = ({
...
+              checked={selectedTypes.trip}
+              onCheckedChange={handleToggleTrip}
             />
-          </HStack>
+          </RStack>
         )}
-        <HStack space={3} alignItems="center">
-          <Text
-            fontSize="lg"
-            fontWeight="bold"
```
andrew-bierman, line 138: we might want bold still

mbonete: Thanks for the piece of feedback! I made that call while refactoring the design as I found the non-bold version closer to the apperance of the original view. I will put back the `bold` and prioritize keeping the HTML/design semantic over visual fidelity.

andrew-bierman: I like your thinking! that's a good thing to be aware of

andrew-bierman:
Possible memory leak in the RStacks in FeedCard? Trying searching with search bar and see, super laggy for me and crashes. I narrowed it down to in here maybe ? But could use second opinion

andrew-bierman:
Dropping a comment as a reminder about memory leak / performance issue when using FeedCard. I think it's the 2 nested Stacks inside the card for some reason, but not 100%. Does not happen with native base weirdly.

andrew-bierman:
https://github.com/andrew-bierman/PackRat/issues/521

PR #489, opened 2023-11-12
Refactor Card to use our tamagui wrapper instead of native-base

mbonete:
Uses #487 as base branch, so the PR diff against `andrew_testing` will show changes that are already present there.

Please notify me if you want me to group the changes for all views in a single PR.


mbonete:
To avoid having dependencies in different PRs I am going to use an unified one. I have changed the base of this PR to `feature/tamagui` and I am merging these changes there.

PR #493, opened 2023-11-17
Worked on search functionality

codecraftwt:


andrew-bierman:
What is large commented section at bottom of feed file now?

andrew-bierman:
Can we go back to using RText so we keep all our ui in the packages/ui? We can simplify it to just reexport tamagui Text if that's the issue

andrew-bierman:
Let's sync this branch back up with the base to ensure it works with latest iteration

andrew-bierman:
Don't delete the example files, others may need to reference them

PR #495, opened 2023-11-22
updated

yash1539:


client/queries/destinationQueries.js
```diff
@@ -0,0 +1,9 @@
+// destinationQueries.js
```
andrew-bierman, line 1: did not use proper base branch. these should be in hooks directory like the others on trpc branch

client/queries/destinationQueries.js
```diff
@@ -0,0 +1,9 @@
...
+import { useQuery } from 'react-query';
```
andrew-bierman, line 2: did not use trpc query integration, which was main point of this task

client/queries/destinationQueries.js
```diff
@@ -0,0 +1,9 @@
...
+
+export function useDestinationData(destinationId, type, id) {
+  return useQuery(['destination', destinationId, type, id], async () => {
+    const response = await fetch(`/api/destination?id=${destinationId}&type=${type}&osm_id=${id}`);
```
andrew-bierman, line 6: we use trpc now ...

PR #501, opened 2023-12-06
trip save fixed

aezazali1997:
### Trip creation fixed

Screenshot 2023-12-06 at 10 57 21 PM

relates to : #500


client/screens/trip/createTrip.tsx
```diff
@@ -31,8 +31,10 @@ export default function Trips() {
     (state) => state.search.selectedSearchResult,
   );
 
-  const { latLng, selectedSearch } = useSelector((state) => state.weather);
-
+  const weather = useSelector((state) => state.weather);
```
andrew-bierman, line 34: we are trying to move away from redux to trpc query, so we should update this with new logic

PR #509, opened 2023-12-17
feat: fix add trip issue and wether card issue 

Jaydip-1002:


client/components/trip/createTripModal.tsx
```diff
@@ -116,18 +116,23 @@ export const SaveTripContainer = ({ dateRange }) => {
       endDate,
     };
 
-    // const geoJSON = await trpc.getPhotonDetails.query({
-    //   id: search.properties.osm_id,
-    //   type: search.properties.osm_type,
+    const geoJSON = await trpc.getPhotonDetails.query({
```
andrew-bierman, line 119: we should be using a the hook here, not directly making query.

andrew-bierman (approved):

PR #510, opened 2023-12-18
Feat/trip fixes

mushtaq-khan:


client/components/trip/createTripModal.tsx
```diff
@@ -93,8 +93,17 @@ export const SaveTripContainer = ({ dateRange }) => {
   // const [startDate, setStartDate] = useState("");
   // const [endDate, setEndDate] = useState("");
   const [isPublic, setIsPublic] = useState(true);
-
-  // create trip
+  const usePhotonDetailsQuery = () => {
```
andrew-bierman, line 96: We should be doing this in the hooks directory not inline here

server/src/middleware/validators/tripRoutesValidator.ts
```diff
@@ -20,7 +20,8 @@ export const addTrip = z.object({
   start_date: z.string().nonempty(),
   end_date: z.string().nonempty(),
   destination: z.string().nonempty(),
-  geoJSON: z.object({}),
+
+  geoJSON: z.unknown(),
```
andrew-bierman, line 24: its not unknown its an object

client/components/trip/createTripModal.tsx
andrew-bierman: Also we probably don't need to refetch this, i believe its done in the parent component

client/components/destination/index.tsx
```diff
@@ -86,7 +86,7 @@ const DestinationHeader = ({ geoJSON, selectedSearchResult }) => {
  * @param {Object} geoJSON - The GeoJSON data.
  * @return {JSX.Element|null} The WeatherCard component if weatherObject and weatherWeek are truthy, otherwise null.
  */
-const WeatherData = ({ geoJSON }) => {
+export const WeatherData = ({ geoJSON }) => {
```
andrew-bierman, line 89: this is for destinations, we should be able to just get the weather card already used working

andrew-bierman (approved):

PR #517, opened 2023-12-21
Fix/routing trip pack

andrew-bierman:
- [ ] If successfully created pack/trip, we should send to the details page with the id. Not just the overall /packs or /trips page

### DIFF
--- a/client/components/pack/AddPack.tsx
+++ b/client/components/pack/AddPack.tsx
@@ -3,20 +3,22 @@ import { Box, Input, Button, Text, Select, CheckIcon } from 'native-base';
 
 // import useAddPack from "../../hooks/useAddPack";
 import { addPack } from '../../store/packsStore';
-import { useState } from 'react';
+import {  useState } from 'react';
 // import { useAuth } from "../../auth/provider";
 import { useSelector, useDispatch } from 'react-redux';
 import { CustomModal } from '../modal';
 import useTheme from '../../hooks/useTheme';
 import useCustomStyles from '~/hooks/useCustomStyles';
 import { useAddNewPack } from '~/hooks/packs';
+import { useRouter } from 'expo-router';
 
-export const AddPack = () => {
+export const AddPack = ({isCreatingTrip}) => {
   const { enableDarkMode, enableLightMode, isDark, isLight, currentTheme } =
     useTheme();
   const styles = useCustomStyles(loadStyles);
 
   const dispatch = useDispatch();
+  const router = useRouter();
 
   const [name, setName] = useState('');
   const [isPublic, setIsPublic] = useState(false);
@@ -24,13 +26,11 @@ export const AddPack = () => {
   // const { addPack } = useAddPack();
   // const { user } = useAuth();
   const user = useSelector((state) => state.auth.user);
-
   const isLoading = useSelector((state) => state.packs.isLoading);
-
-  const error = useSelector((state) => state.packs.error);
-
-  const isError = error !== null;
-  const { addNewPack } = useAddNewPack();
+  const { addNewPack,isSuccess,isError } = useAddNewPack();
+    if (isSuccess && !isCreatingTrip ) {
+      router.push('/packs');
+    }
   /**
    * Handles the addition of a pack.
    *
@@ -105,7 +105,7 @@ export const AddPack = () => {
   );
 };
 
-export const AddPackContainer = () => {
+export const AddPackContainer = ({isCreatingTrip}) => {
   const [isOpen, setIsOpen] = useState(false);
   return (
     <CustomModal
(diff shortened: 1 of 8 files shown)


client/components/pack/AddPack.tsx
```diff
@@ -3,34 +3,34 @@ import { Box, Input, Button, Text, Select, CheckIcon } from 'native-base';
...
-
   const isLoading = useSelector((state) => state.packs.isLoading);
-
-  const error = useSelector((state) => state.packs.error);
-
-  const isError = error !== null;
-  const { addNewPack } = useAddNewPack();
+  const { addNewPack,isSuccess,isError } = useAddNewPack();
+    if (isSuccess && !isCreatingTrip ) {
+      router.push('/packs');
```
andrew-bierman, line 32: we should get the id from response and direct to there. double check but i think it will be singular /pack/:packId for the route

client/components/pack/AddPack.tsx
```diff
@@ -3,34 +3,34 @@ import { Box, Input, Button, Text, Select, CheckIcon } from 'native-base';
...
 // import { useAuth } from "../../auth/provider";
 import { useSelector, useDispatch } from 'react-redux';
 import { CustomModal } from '../modal';
 import useTheme from '../../hooks/useTheme';
 import useCustomStyles from '~/hooks/useCustomStyles';
 import { useAddNewPack } from '~/hooks/packs';
+import { useRouter } from 'expo-router';
 
-export const AddPack = () => {
+export const AddPack = ({isCreatingTrip}) => {
```
andrew-bierman, line 15: lets default to false here

client/hooks/trips/useAddTrip.ts
```diff
@@ -1,15 +1,19 @@
...
 export const useAddTrip = () => {
+  const router = useRouter();
   const utils = queryTrpc.useUtils();
   const mutation = queryTrpc.addTrip.useMutation();
 
   const addTrip = (newTrip) => {
     mutation.mutate(newTrip, {
       onSuccess: () => {
         utils.getTrips.invalidate();
+        router.push('/trips');
```
andrew-bierman, line 13: same comment about id as above, although i wonder if this hook is the best place to be routing in 🤔 fine for now, might be only option without fully refactoring the hook

server/src/models/userModel.ts
```diff
@@ -1,6 +1,6 @@
 import mongoose, { Schema, type Document, type Model } from 'mongoose';
 import myDB from './dbConnection';
-import bycrypt from 'bcrypt';
+import bcrypt from 'bcrypt';
```
andrew-bierman, line 3: dang lol, is this all we need to fix the hash issue? can we revert your changes below if this is working ?

shahzi113awan: No, I think the problem is in siginup
Look I just  signed up now and my user data in db looks like 

_id
6584f96987973d18cbe82a52
name
"shezzii"
password"12345678"
...
}
For now I just bypassed this by just commenting preSave hook in model

andrew-bierman: It's recent which is strange. I'm seeing sign ups from a few days ago properly hashed

server/src/models/userModel.ts
```diff
@@ -117,6 +115,7 @@ UserSchema.statics.alreadyLogin = async function (
   email: string,
 ): Promise<void> {
   const user = await User.findOne({ email });
+  console.log('user', user);
```
andrew-bierman, line 118: clean up

server/src/models/userModel.ts
```diff
@@ -103,11 +103,9 @@ UserSchema.statics.findByCredentials = async function ({
   password: string;
 }): Promise<IUser> {
   const user = await User.findOne({ email });
-
   if (!user) throw new Error('Unable to login');
-
-  const isMatch = await bycrypt.compare(password, user.password);
-
+// console.log('user', password, user.password);
```
andrew-bierman, line 107: clean up

client/store/authStore.ts
```diff
@@ -31,7 +31,7 @@ export const signUp = createAsyncThunk(
         email,
         password,
       });
-      return user;
+      // return user;
```
andrew-bierman, line 34: clean up, we need to do one or the other but not call mutate twice (not your fault it was weird how it was set up)

client/components/trip/createTripModal.tsx
```diff
@@ -139,7 +139,10 @@ export const SaveTripContainer = ({ dateRange }) => {
     addTrip(data);
     setIsSaveModalOpen(!isSaveModalOpen);
   };
-
+  if (isSuccess  && response) {
+    console.log('response ->', response);
```
andrew-bierman, line 143: clean up

andrew-bierman:
Approved, just clean up some of those logs and touched files if possible

PR #524, opened 2023-12-24
Update database model schemas with Drizzle ORM

sheharyarahmad842:


sheharyarahmad842:
I have updated the schemas of all the database models using Drizzle ORM as assigned. Plz let me know if there are potential issues so I can fix them immediately.

raghavyuva:
this seems pretty good!, i have gone through the code and merging it to the main pr so that we can take it from there while testing the api's

PR #525, opened 2023-12-26
Convert disconnect codes to drizzle and fix drizzle-kit studio issue related to schemas

sheharyarahmad842:


sheharyarahmad842:
Tried converting disconnect codes. Kindly review and tell if this approach is fine. Plz let me know if something is wrong.

PR #535, opened 2023-12-30
moving SearchInput logic to custom hook, adding to component

andrew-bierman:


andrew-bierman:
https://github.com/andrew-bierman/PackRat/issues/529

PR #537, opened 2024-01-01
About screen

fshah21:


andrew-bierman:
Put these functions into a useAboutLogic or something similar hook so the syntax aligns with the others

PR #538, opened 2024-01-01
fix: register screen bug

taronaleksanian:


andrew-bierman (approved):
nice job, this seems to fix errors on login/register screens

PR #539, opened 2024-01-01
Convert template related services from prisma to drizzle orm

sheharyarahmad842:


sheharyarahmad842:
I have tried converting template and trip related services. Please let me know if there are issues in it.

server/src/controllers/trip/deleteTrip.ts
```diff
@@ -1,7 +1,7 @@
 import { publicProcedure } from '../../trpc';
 import { UnableToDeleteTripError } from '../../helpers/errors';
 import * as validator from '../../middleware/validators/index';
-import { Trip } from '../../drizzle/methods/trip';
+import { Trip } from '../../drizzle/methods/Trip';
```
raghavyuva, line 4: Why this is Trip any errors to change this?

server/src/controllers/trip/editTrip.ts
```diff
@@ -34,7 +34,7 @@ export function editTripRoute() {
   return publicProcedure.input(validator.editTrip).mutation(async (opts) => {
     const { id, ...rest } = opts.input;
     const tripClass = new Trip();
-    const trip = await tripClass.update(rest,id);
-    return trip 
+    const trip = await tripClass.update(rest, id);
```
raghavyuva, line 37: Not sure why you are touching trip related services in this PR

raghavyuva (changes requested):
Refer creating db instance in main pr

server/src/controllers/trip/deleteTrip.ts
sheharyarahmad842: The name of the methods were capitalized so I thought to change "trip" to "Trip". But I'll rename it to trip

server/src/controllers/trip/editTrip.ts
sheharyarahmad842: Trip and template related services were left so I tried changing them. I asked from Andrew about it.

raghavyuva:
> I have tried converting template and trip related services. Please let me know if there are issues in it.

trip is already converted,

sheharyarahmad842:
I should have pulled the latest code. Now plz tell me what should I do. I don't have much experience with git and github. How can we merge this PR because there can be conflicts.

raghavyuva:
The best thing to do is soft reset head by one commit and make sure to commit back by adding only those files/codes which are relevant to the pr

PR #542, opened 2024-01-02
items screen

fshah21:
Still working on this.


andrew-bierman:
is this still in progress or good to merge ?

andrew-bierman:
We are still rendering the old item screen from (app) directory also FYI - we should get that fixed with this PR

andrew-bierman (changes requested):
Fix to make proper items screen display

andrew-bierman (approved):
approving so you don't lose work, we still need to render this screen in file routing though, none of this is being applied currently due

PR #546, opened 2024-01-04
refactor: login screen

taronaleksanian:


andrew-bierman (approved):
nice work, unsure if we want google auth to have it's own folder. will see how it compares with the register one

PR #550, opened 2024-01-05
refactor: refactor navigation components

taronaleksanian:


andrew-bierman:
This makes me wonder if we should have a router hook and import from there, instead of using direct expo router import, in case we change to solito in the future. Just an idea

taronaleksanian:
> solito

Yeah, I think so, I will change it that way.

taronaleksanian:
> This makes me wonder if we should have a router hook and import from there, instead of using direct expo router import, in case we change to solito in the future. Just an idea

done

andrew-bierman (approved):

PR #560, opened 2024-01-07
refactor Progress component

yangTauTauAjah:
This is my first commit, haven't done too much tho, just ensuring everything works properly before I start doing any further changes

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,38 @@
+module.exports = {
+    "env": {
+        "browser": true,
+        "es2021": true,
+        "node": true
+    },
+    "extends": [
+        "eslint:recommended",
+        "plugin:@typescript-eslint/recommended",
+        "plugin:react/recommended"
+    ],
+    "overrides": [
+        {
+            "env": {
+                "node": true
+            },
+            "files": [
+                ".eslintrc.{js,cjs}"
+            ],
+            "parserOptions": {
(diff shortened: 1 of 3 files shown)


.eslintrc.js
```diff
@@ -0,0 +1,38 @@
+module.exports = {
```
andrew-bierman, line 1: do we need this file?

client/components/progress/index.tsx
```diff
@@ -1,41 +1,14 @@
...
-  }, [reduxTargetValue]);
-
-  useEffect(() => {
-    if (localCurrentValue < reduxTargetValue && operationId !== 0) {
-      setLocalCurrentValue((prevValue) => prevValue + 1);
-    }
-  }, [localCurrentValue, reduxTargetValue, operationId]);
+  const {
+    localCurrentValue,
+    currentTheme
```
andrew-bierman, line 8: the useTheme logic can actually stay in the component, only because we will be refactoring these component to use tamagui theme here soon. No worries, i did not explain that detail well. nice job :)

.eslintrc.js
yangTauTauAjah: > do we need this file?

i guess so, since we have tsconfig.json renamed to tsconfig.base.json, this will tell eslint to read that config file for linting ts files. or do you have any advice?

PR #561, opened 2024-01-07
move map preview api call to backend

mikib0:


andrew-bierman:
Nice job with this, can you just update your logic to use trpc instead of direct HTTP request (you can keep your logic for http, just add corresponding trpc route similar to other controllers and hook to make the fetch). We have moved all our frontend requests to trpc since you were last involved so no worries, ping me if you need assistance. Happy to help out.

andrew-bierman (approved):
approving current version as i don't want this work to be messed up after restructuring

PR #562, opened 2024-01-08
Feat/implementing tests for trpc routes

andrew-bierman:
WIP

The goal here is just to get some basic coverage of our trpc routes, so that we can better monitor breaking changes - especially as it relates to the upcoming integration of drizzle orm.

I have provided a brief example of how we can use zod-mock to move quickly to get these tests working. This is set up with a /tests directory. I think we can keep the root test logic here, but it might be better to colocate the tests within the proper router/controller. If we add sub directories for these, it might be easier to follow the logic. Open to either. 

### Task List 
- [ ] Create 'green' (success) tests for all routes in trpcRouter.ts file


andrew-bierman:
The goal here is just to get some basic coverage of our trpc routes, so that we can better monitor breaking changes - especially as it relates to the upcoming integration of drizzle orm.

PR #564, opened 2024-01-08
Refactor/modals

taronaleksanian:


andrew-bierman:
I'm no longer seeing the X in top right corner of our modals, i don't think its due to this PR but while you are looking at these modals could you add that back if easy? If not easy we can take it up separately

taronaleksanian:
> I'm no longer seeing the X in top right corner of our modals, i don't think its due to this PR but while you are looking at these modals could you add that back if easy? If not easy we can take it up separately

OK, sure I will review it, and try to fix it ASAP

taronaleksanian:
> I'm no longer seeing the X in top right corner of our modals, i don't think its due to this PR but while you are looking at these modals could you add that back if easy? If not easy we can take it up separately

done

andrew-bierman (approved):
nice job with this :)

PR #567, opened 2024-01-08
refactor: refactor custom button

taronaleksanian:


andrew-bierman:
Can we replace custom button with just our normal button from @packrat/ui ? Is it doing anything differently?

taronaleksanian:
> Can we replace custom button with just our normal button from @packrat/ui ? Is it doing anything differently?

It has its styles, I think we can have Button variants later. Probably with tagmaui theme config, I just tried to keep the version that we have now, and collect all the buttons we have now, so later creating variants will be easier

taronaleksanian:
> Can we replace custom button with just our normal button from @packrat/ui ? Is it doing anything differently?

actually, you are right, I though it was created as a button variant, but after reviewing it carefully I saw that there is nothing about colors or sizes

andrew-bierman (approved):
good work, we should at some point refactor so file structure of button matches other UI components. But that can be a later task this is fine for now

PR #568, opened 2024-01-08
refactor: move multistep form business logic

taronaleksanian:


client/components/MultistepForm/MultistepForm.tsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+import { View, Text, TouchableOpacity } from 'react-native';
+import useCustomStyles from '~/hooks/useCustomStyles';
+import { useMultiStepForm } from './useMultiStepForm';
+import { ProgressBar } from './ProgressBar';
+import { Sidebar } from './Sidebar';
+
+// TODO move to packages, we need to inject styles in this component, need to think what is the best way to do it
```
andrew-bierman, line 8: that is a great point, this would be really cool as a reusable ui component. For styling, i say we just rewrite to tamagui to inherit theme styles, might be easier solution and match other ui.

andrew-bierman (approved):

PR #576, opened 2024-01-10
Refactor/types

uros-simeunovic:


client/components/map/WebMap.web.tsx
```diff
@@ -32,7 +32,7 @@ import {
   multiPolygonBounds,
 } from '../../utils/mapFunctions';
 import MapButtonsOverlay from './MapButtonsOverlay';
-import { saveFile } from '../../utils/fileSaver/fileSaver';
+import { saveFile } from '../../utils/fileSaver/fileSaver.native';
```
andrew-bierman, line 35: we should not do this, react native resolves the .native and .web extensions per platform so it doesn't get compiled.

andrew-bierman: to fix the type error we need to make these into folders with index and custom exports I believe

andrew-bierman:
dispatcher is null error on /trip/create page (createTrip)

packages/app/components/item/searchItem.tsx
```diff
@@ -52,19 +52,17 @@ export const SearchItem: React.FC<Props> = ({ onSelect, placeholder }) => {
...
             }}
             placeholder={placeholder ?? 'Type here to search'}
             value={searchString}
             fontSize={14}
-            width={'100%'}
+            width="100%"
             borderRadius={4}
-            paddingVertical={16}
-            paddingHorizontal={8}
+            padding="16px 8px"
```
andrew-bierman, line 65: we are moving away from px sized styles as they work on web but not on native also FYI

andrew-bierman (approved):

andrew-bierman (approved):

PR #589, opened 2024-01-11
Fix/add types and interfaces in components

jitprosen360:
Component files adding types and interfaces


andrew-bierman (changes requested):
please rebase or copy changes to latest file structure

client/app.example.json
andrew-bierman, line 1: don't delete example file

jitprosen360: solved

andrew-bierman:
these should be going into the /packages/app directory now, not /client directory. copy these changes to a clean branch off of andrew_testing please

PR #590, opened 2024-01-11
Fix and test trip related routes

sheharyarahmad842:


sheharyarahmad842:
I have comprehensively tested trip related routes. If still issues are encountered while checking plz let me know.

andrew-bierman (approved):
nice work

PR #594, opened 2024-01-12
Refactor/item

yangTauTauAjah:


packages/app/components/item/AddItem.tsx
```diff
@@ -1,98 +1,43 @@
...
 import { View } from 'react-native';
 import { useDispatch, useSelector } from 'react-redux';
 import { ItemForm } from './ItemForm'; // assuming you moved the form related code to a separate component
 import { useAddPackItem } from 'app/hooks/packs/useAddPackItem';
 import { useEditPackItem } from 'app/hooks/packs/useEditPackItem';
+import useAddItem from 'app/hooks/item/useAddItem';
 
 export const AddItem = ({
-  _id,
+  // _id,
```
andrew-bierman, line 10: why are we disabling these just curious ?

yangTauTauAjah: it seems that we're not gonna use that variabel yet, unless we're enabling line 70 on useAddItem.tsx

PR #595, opened 2024-01-12
trpc route for map preview

mikib0:


andrew-bierman:
Are you seeing this error when trying to view a preview on frontend ? 

```typescript
No "query"-procedure on path "getMapPreview"
```

mikib0:
True, I see the error on the render deploy, but the same commit works fine locally. Would you be able to try it locally too?

PR #597, opened 2024-01-12
Test item related routes comprehensively

sheharyarahmad842:


sheharyarahmad842:
Plz thoroughly check and test the routes before merging and let me know if there are issues.

andrew-bierman (approved):
nice work with these

PR #602, opened 2024-01-13
Refactor/card

taronaleksanian:


packages/ui/src/ThreeDotsMenu/ThreeDotsMenu.tsx
```diff
@@ -0,0 +1,41 @@
+import { MoreHorizontal } from '@tamagui/lucide-icons';
```
andrew-bierman, line 1: this would be a cool spot to use Zeego library

andrew-bierman (approved):
nice work with this

PR #606, opened 2024-01-14
move password reset page to (auth)

gauthamdas:
- moved `/password-reset` -> `(auth)`
- added a fix: to catch sendGrid email exception


andrew-bierman (approved):
good catch on this one

PR #608, opened 2024-01-14
Custom hooks

shewit-gh:


packages/app/components/select/index.tsx
```diff
@@ -1,6 +1,6 @@
-import { Check, ChevronDown, ChevronUp } from '@tamagui/lucide-icons';
+import { ChevronDown, ChevronUp } from '@tamagui/lucide-icons';
```
andrew-bierman, line 1: this should be in UI folder if not already

packages/app/hooks/select/useSelect.ts
```diff
@@ -0,0 +1,55 @@
+import { useState } from 'react';
```
andrew-bierman, line 1: this hook might be leftover logic from the docs or something, i don't think we need

server/nodemon.json
```diff
@@ -2,5 +2,5 @@
     "watch": ["src"],
     "ext": ".ts,.js",
     "ignore": ["node_modules"],
-    "exec": "tsx src/index.ts"
```
andrew-bierman, line 5: tsx might be faster, i added when you said ti was running slow on your machine

andrew-bierman:
for the useCustomStyles stuff - you can actually leave these in the components for the time being. Once we clean up ui, these will not be needed as our theming will be handled via tamagui

PR #610, opened 2024-01-14
[Refactor] Added new types

uros-simeunovic:


andrew-bierman:
Should we keep the types folder in the server and then import them to frontend ? we already have server installed as package for trpc. And it looks like you based these off of the backend models

uros-simeunovic:
In server/src/models only userModel has user interface. I can make interface for every model there and i will delete types from packages.

andrew-bierman:
Ah ok. Thought there were more. Don't worry about that, we will get types from drizzle PR soon. We can use these in frontend for now actually is probably easiest approach. Let's go through and fix other type errors in /packages/app now

andrew-bierman (approved):

PR #622, opened 2024-01-18
Fix/types interface add components screens

jitprosen360:
type and interface add in components


andrew-bierman:
the carousel buttons are broken with these changes

PR #625, opened 2024-01-19
Fix and test pack/favorite related routes

sheharyarahmad842:


sheharyarahmad842:
Tried refactoring, testing and fixing issues in pack and favorite related routes.

PR #632, opened 2024-01-21
Tanstack table integrations

shahzi113awan:


packages/ui/src/DataTable/index.tsx
```diff
@@ -0,0 +1,31 @@
...
+  const table = useReactTable({
+    data,
+    columns,
+    getCoreRowModel: getCoreRowModel(),
+    enableColumnResizing:true,
+    columnResizeMode:  'onChange'
+  });
+
+  return (
+    <div className="p-2">
```
andrew-bierman, line 23: does this need to be a div ? we should probably use a tamagui stack or box here so it works on native

PR #644, opened 2024-01-25
feat: expo drawer + tabs

imrizwan:


andrew-bierman:
Nice work with this, leaving some small comments so web versio doesn't crash

packages/app/components/navigation/Navigation.tsx
```diff
@@ -1,24 +1,47 @@
...
 import useCustomStyles from 'app/hooks/useCustomStyles';
 import { useIsMobileView } from 'app/hooks/common';
 import { useNavigate } from 'app/hooks/navigation';
 import { useAuthUser } from 'app/hooks/user/useAuthUser';
 import { NavigationList } from './NavigationList';
 import { Button } from 'tamagui';
 import SVGLogoComponent from '../../components/logo';
-import { Platform } from 'react-native';
-import { Image } from 'solito/image';
+import { Drawer } from 'expo-router/drawer';
```
andrew-bierman, line 11: I believe we will need to extract this into a drawer.native.tsx and drawer.web.tsx files so this doesn't crash on next js side

packages/app/components/navigation/Navigation.tsx
```diff
@@ -30,24 +53,14 @@ export const Navigation = () => {
             variant="outlined"
             chromeless
           >
-            {/* FIX ME */}
             <View style={styles.logoWrapper}>
-              {Platform.OS === 'web' ? (
```
andrew-bierman, line 35: we should not remove this, we still need to fix the svg logo on next js side

PR #647, opened 2024-01-27
Next prefetch - Effort Estimation 🐎 

raghavyuva:
# Next Pre-fetch 🐎 

### To the People Who wants to know about pre-fetching
Prefetching is the process of loading data in advance to speed up its future access. In the context of web development, server-side helpers can prefetch data that is needed to render a page. This means when the page is requested, much of the data it needs is already loaded and can be immediately used to generate the HTML. This approach reduces the time taken to load the data on the client side since the server has already done the heavy lifting.

### The story
Since we have moved our all the logics from expo web to nextjs in the last few days, we are trying to make use of better nextjs functionalities available out there. While keeping our app logics up to date.

### Current Stage ( Effort Estimation 😄 )
We are just testing out how much effort it might take to implement this feature for all the pages and routes inside the packrat next.

1. Do some configuration for prefetch related stuffs on client side,
2. Test for individual route / api for this approach making use of the documentation
3. Estimate the effort and time it will require for entire codebase integration (nextjs)
4. Do repeat for other routes/api's if it is feasible

### DIFF
--- a/apps/next/pages/pack/[id].tsx
+++ b/apps/next/pages/pack/[id].tsx
@@ -1,10 +1,15 @@
 import React from 'react';
 import { PackDetails } from 'app/components/pack/PackDetails';
 import { AuthWrapper } from 'auth/authWrapper';
+import { GetServerSidePropsContext, InferGetServerSidePropsType } from 'next';
+import { helpers } from 'prefetchConf';
 
 // export const runtime = 'experimental-edge'
 
-function PackScreen() {
+function PackScreen(
+  props: InferGetServerSidePropsType<typeof getServerSideProps>,
+) {
+  const { initialPackData } = props;
   return (
     <>
       <PackDetails />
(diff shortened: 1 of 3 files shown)


reubence:
hi can i take this up? /attempt #647

oliverqx:
Hello,

I'm very familiar with NextJS app and pages router. I've migrated quite a lot of apps from react to nextjs and from the pages router to the app router. I'd love to contribute with the process 

Will the discussion take place here?
/attempt #647

oliverqx:
cc: @raghavyuva

raghavyuva:
> cc: @raghavyuva 

Take this up if you want to ,

MayorChristopher:
Is it still open?, if I may work on it

oliverqx:
I've been assigned I'm just waiting on answers to some questions I've made

andrew-bierman:
> Hello,
> 
> 
> 
> I'm very familiar with NextJS app and pages router. I've migrated quite a lot of apps from react to nextjs and from the pages router to the app router. I'd love to contribute with the process, just have a couple of questions:
> 
> 
> 
> 1. What do you mean by client side prefetching? As in route prefetching?
> 
> 2. I see app/nextjs is using the pages router, do you plan on migrating to the app router?
> 
> 3. Where will the nextjs app live? Will you be using vercel?

1. We should prefetch our tRPC queries server side on next js. Allowing our tRPC query cache to have accurate data before it arrives to the user. Many components are reused between next and expo native app, so in next routes the idea is to populate the cache in advance, but on expo side it will fetch client side. Client side fetching is currently implemented, and using cached value instead of refetching is supported with trpc query. 
2. Not at this time, when more libraries support it we will eventually. 
3. Cloudflare

oliverqx:
This past two days (monday,tuesday), I've been looking into trpc ssr/isr/isg support and I think ive a pretty good idea of how this would be done. 

I'm running some tests today and will have a better understanding of the time needed to complete this task afterwards

oliverqx:
sorry I couldnt keep going. I've been crazy busy this past week and it was surprisingly difficult to enable ssr with trpc

PR #649, opened 2024-01-28
Refactor/auth

taronaleksanian:


andrew-bierman:
Are you able to sign in / sign out on this branch? It doesn't seem to be sending sign in mutation on my side.

taronaleksanian:
> Are you able to sign in / sign out on this branch? It doesn't seem to be sending sign in mutation on my side.

Yeah it works, on my side, I forgot to mention, that we need to add .env in the next js app folder, with the folowing values

NEXT_PUBLIC_APP=next
NEXT_PUBLIC_GOOGLE_ID =993694750638-876nvl8hfjfv6tqiqc1c4tmubvf8chpp.apps.googleusercontent.com

packages/app/utils/storage.ts
```diff
@@ -0,0 +1,67 @@
+import EventEmitter from 'events';
+import * as SecureStore from 'expo-secure-store';
+
+const isNextApp = process.env.NEXT_PUBLIC_APP === 'next';
```
andrew-bierman, line 4: I think a cleaner way to handle would be to have a storage.native.ts file / storage.web.ts file. The logic can be separated across the two, and react native web automatically uses the proper file.

taronaleksanian: Exactly, I will change it

andrew-bierman:
somehow there are some unrelated commits that made it in here. let's cherry pick to only get the auth related stuff in this PR

apps/next/auth/authWrapper.tsx
andrew-bierman, line 1: We should probably move this auth wrapper into that app/auth folder where you had the other stuff if its going to be this similar that we can reuse

taronaleksanian: I see AuthLoader is implemented again in the tauri app, but AuthLoader and AuthWrapper are different components, Authloader is that shared logic that we have in the app folder, and thought it's better to separate it's logic within the apps. Anyway agreed we can unify it too, so we also have unified router in the app workspace, so I will change it

andrew-bierman: fair point, I guess my thinking at this stage is try to keep as much logic and ui in the packages/app as possible, and really only handle core platform specific config stuff and file based routing in the apps folder. but agree its definitely a bit of a grey area

PR #653, opened 2024-01-31
refactor: refactor search input

taronaleksanian:


packages/app/screens/trip/createTrip.tsx
```diff
@@ -60,131 +61,12 @@ export default function Trips() {
...
-  // }, [parksObject]);
-
-  const { data: photonDetails } = useGetPhotonDetails({
-    properties: {
-      osm_id: searchResult.properties?.osm_id,
-      osm_type: searchResult.properties?.osm_type,
-    },
-  });
-
-  const steps = [
```
andrew-bierman, line 86: this is for native side where it appears as a multi step form rather than one large scrollable page

andrew-bierman:
the actual search logic seems good so nice work there, gonna drop some comments on some things i noticed that might have broken (could have been before this PR also)

andrew-bierman:
- [ ] On dashboard, searching for a location works but upon selection we are not fetching the photon details from backend and displaying the map / weather

andrew-bierman:
- [ ] on create trip, i get this error when we try to check if the object is empty "TypeError: Cannot convert undefined or null to object"

andrew-bierman (approved):
nice work with this

PR #662, opened 2024-02-06
Refactor/auth new

taronaleksanian:


apps/expo/app/(drawer)/_layout.tsx
```diff
@@ -1,9 +1,14 @@
...
+import { Redirect, Stack } from 'expo-router';
+import { AuthLoader } from 'app/auth/AuthLoader';
+import { Text } from '@packrat/ui';
 
 export default function AppLayout() {
-  const { session } = useSession();
-  useProtectedRoute(session);
-  return <Navigation />;
+  return (
+    <AuthLoader
```
andrew-bierman, line 7: Not sure why but this breaks native for some reason, despite logic being quite similar to what we had before

packages/app/auth/AuthLoader.tsx
```diff
@@ -0,0 +1,19 @@
+import { useUserQuery } from './hooks';
+
+export const AuthLoader = ({
+  children,
+  loadingElement,
+  unauthorizedElement,
+}) => {
+  const { user, isLoading } = useUserQuery();
```
andrew-bierman, line 8: i think the data coming back from here is what is causing the issue on native side. works fine on web but i'm having issues on ios

packages/app/auth/hooks/useUser.ts
```diff
@@ -0,0 +1,14 @@
+import { queryTrpc } from 'app/trpc';
+import { useStorage } from 'app/hooks/storage/useStorage';
+
+export const useUserQuery = () => {
+  const [[isLoading, user]] = useStorage('session');
```
andrew-bierman, line 5: this data comes back differently from secure storage versus local storage on web. we might want to make this a bit easier to follow too

andrew-bierman:
Upon further reading, expo secure store can only store strings. Not objects. Meaning our current 'session' part of storage works on web but not native. We should refactor to use the token only for authentication loader, not entire user object. 

Then for the useUserQuery that should just be retrieving user object from trpc query. That way it will use persisted cache if available or fetch. 

Open to any alternatives but this seemed to be easiest refactor i could see

PR #665, opened 2024-02-08
Refactor ready for review, all the items listed in both components and screens are tested and is complete.

pinocchio-life-like:
-trip,
-About,
-Appearance,
-landing_page,
-packtable,
-chat,
-map,
-Loader.tsx,
-PackOptions
-inventory (no need for refactoring) 
and more


packages/app/components/pack_table/PackTableStyles.tsx
andrew-bierman, line 1: if you are moving styles to a file make them a .style extension so its easier to follow

packages/app/screens/trip/useWeather.tsx
andrew-bierman, line 1: check logic for this trip section, this is not working on my side. pretty sure we moved away from redux store for weather.

PR #666, opened 2024-02-08
Refactor - redux usage

uros-simeunovic:


andrew-bierman:
a couple things might have been missed and need to go to custom hooks, will drop a few I noticed below

andrew-bierman:
- [ ] changePackStatus
- [ ] addPackItem
- [ ] addTrail 
- [ ] addPark
- [ ] global items (what is useItems.old.ts ?)
- [ ] ....

this is a good one to do but let's search through the store files you are deleting and make sure everything getting removed has a custom hook using trpc query that is being used to replace it

PR #670, opened 2024-02-09
Trip component refactor functionality checked.

pinocchio-life-like:


packages/app/screens/trip/TripDetails.tsx
```diff
@@ -1,52 +1,45 @@
...
   const styles = useCustomStyles(loadStyles);
-  console.log('trip detail');
-  const dispatch = useDispatch();
   const [tripId] = useParam('tripId');
 
   // console.log("🚀 ~ file: TripDetails.js:34 ~ TripDetails ~ tripId:", tripId)
   const { data, isLoading, error, refetch, isOwner, isError } =
     useFetchSingleTrip(tripId);
 
-  const { weatherObject, weatherWeek } = useTripWeather(data);
```
andrew-bierman, line 37: why remove this ?

packages/app/screens/trip/TripDetails.tsx
```diff
@@ -1,52 +1,45 @@
...
-
   const link = `${CLIENT_URL}/trip/${tripId}`;
 
   // useEffect(() => {
   //   if (!tripId) return;
   //   dispatch(fetchSingleTrip(tripId));
   // }, [dispatch, tripId]);
-  const states = useSelector((state: RootState) => state);
```
andrew-bierman, line 45: Why remove type ?

packages/app/screens/trip/TripDetails.tsx
pinocchio-life-like: > why remove this ?

Meant to manage it inside useWeather hook. to create its own hook, instead of inside app/hooks/signleTrips hook component. now I think is of no use, so I revoked.

packages/app/screens/trip/TripDetails.tsx
pinocchio-life-like: > Why remove type ?

Opted for brevity, but I see value in explicit typing. Will revert for consistency.

PR #677, opened 2024-02-11
Refactor/global state

taronaleksanian:


andrew-bierman:
nice work with this, but i'm noticing by removing some dispatch calls some of our CRUD is breaking. I think a few of the places we had dispatch and custom hooks implemented is due to the custom hooks not working possibly. trying to gather list

packages/app/global-state/progressStore/progressStore.ts
```diff
@@ -0,0 +1,66 @@
+import { atom } from 'jotai';
+import { produce } from 'immer';
+import { useReducerAtom } from 'jotai/utils';
+import * as Types from './types';
+
+import EventEmitter from 'events';
```
andrew-bierman, line 6: just curios why do we need these events and emitter logic in here ?

andrew-bierman:
- [ ] double check favoritesStore conversion getting some errors on my side

packages/app/global-state/progressStore/progressStore.ts
taronaleksanian: With jotai we can't dispatch actions for reducer atoms outside of the component, so I ended up creating this logic so we can do it, for the progress bar state. It seems we don't use the progress bar now, but it is there, so I think it's better to keep its support.

andrew-bierman (approved):
nice job with this

PR #679, opened 2024-02-12
Refactor/forms

taronaleksanian:


andrew-bierman:
nice this is exactly what i was thinking for this feature

packages/app/auth/hooks/useLogin.ts
```diff
@@ -8,20 +8,13 @@ interface UserForm {
...
 interface UseLoginReturn {
-  form: UseFormReturn<UserForm>;
   handleLogin: (data: UserForm) => void;
   hasError: boolean;
 }
 
 export const useLogin = (): UseLoginReturn => {
   const { mutateAsync: signIn, error } = queryTrpc.signIn.useMutation();
   const sessionSignIn = useSessionSignIn();
-  const form = useForm({
```
andrew-bierman, line 19: do we lose defaultValues going this route? sorry if i missed where you moved them to

taronaleksanian: It works perfectly even if you don't add default values, so I thought don't add them if we don't need them

packages/ui/src/form/components/Select.tsx
```diff
@@ -0,0 +1,10 @@
+import { LmSelectRhf, type LmSelectRhfProps } from '../lib';
+
+interface Props
+  extends LmSelectRhfProps<{ label: string | number; value: any }> {}
+
+// TODO change the name to "Input" after handling tamagui all components export,
```
andrew-bierman, line 6: ah ok i see

packages/ui/babel.config.js
```diff
@@ -0,0 +1,4 @@
+module.exports = withTamagui({
```
andrew-bierman, line 1: this might need to go into expo/next folder i think, not sure the ui runs babel on build

taronaleksanian: I bet you are right, and just fixed it

andrew-bierman (approved):
really nice work on this one Taron :)

PR #690, opened 2024-02-15
fix library import inside components

pinocchio-life-like:
I have also added some refactoring inside screens, and I will continue doing that.

### DIFF
--- a/packages/app/screens/user/useUserData.tsx
+++ b/packages/app/screens/user/useUserData.tsx
@@ -1,0 +1,21 @@
+import { useState, useEffect } from 'react';
+
+export const useUserData = (user, isLoading) => {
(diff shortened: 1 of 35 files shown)


packages/app/screens/user/useUserData.tsx
```diff
@@ -0,0 +1,21 @@
+import { useState, useEffect } from 'react';
+
+export const useUserData = (user, isLoading) => {
+  const [profileImage, setProfileImage] = useState(null);
+  const [userRealName, setUserRealName] = useState(null);
+  const [username, setUsername] = useState(null);
+
+  useEffect(() => {
```
andrew-bierman, line 8: why useEffect here?

PR #692, opened 2024-02-16
Fixed User Data Cards StyleSheet in Profile

MuhammadHassan03:
Fixed User Data Cards StyleSheet in Profile


apps/expo/env.example
andrew-bierman, line 1: don't delete example files

packages/app/components/DuplicateIcon/index.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 import { Pressable } from 'react-native';
 import { MaterialIcons } from '@expo/vector-icons';
-import { useRouter } from 'app/hooks/router';
+import { useRouter } from 'expo-router';
```
andrew-bierman, line 4: should be coming from app/hooks. we don't use expo router on all platforms

apps/expo/env.example
MuhammadHassan03: sure, i have included this....... and going to pull request again

packages/app/components/DuplicateIcon/index.tsx
MuhammadHassan03: yeah, i got it and have changed it

PR #700, opened 2024-02-18
fix: trip page param issue

kidussintayehu:


andrew-bierman:
Issue still persists on this branch. To re-create:
- Go to /trip/create
- Select a location
- then change the selected pack. Params clear out and we lose state

tibeb-dev:
Fixed now. check it out!

PR #704, opened 2024-02-20
fix for createTrip component dropping state

pinocchio-life-like:
Fix for createTrip Component dropping the already established states when selecting different options on the page.


andrew-bierman (approved):
good catch on this

PR #718, opened 2024-02-25
Feat/pull to refresh with refreshcontrol

imrizwan:


packages/app/components/itemtable/itemTable.tsx
```diff
@@ -70,25 +73,81 @@ export const ItemsTable = ({
...
+      </Text>,
+      <Text
+        style={styles.rowText}
+        key="category"
+      >{`${category?.name || type}`}</Text>,
+      <View
+        key="viewContainer"
+        style={{ flex: 1, justifyContent: 'center', alignItems: 'flex-end' }}
+      >
+        <DropdownMenu.Root>
```
andrew-bierman, line 106: We should pull this into another component to its easier to follow. same file is fine. 

also if we are going to use DropDownMenu directly we should add some basic styles to web version so it looks decent. the docs have a good example using dripsy we could apply to tamagui/vanilla stylesheet.

andrew-bierman:
good job here, i think we could use refresh on the details pages (ie pack details) also. but otherwise great work you nailed it. 

gonna drop a couple small cleanup tasks below

andrew-bierman:
- [ ] refetch scroll on pack / trip details pages
- [ ] styling for zeego dropdown on web (see zeego docs, dripsy example)

andrew-bierman:
- [ ] same thing you did on items table should be done on pack table

packages/app/screens/LoginScreen.tsx
```diff
@@ -7,9 +7,6 @@ import {
   RText,
   RIconButton,
   RScrollView,
-  Form,
```
andrew-bierman, line 10: i think force push might have overwritten the custom form changes

andrew-bierman:
this looks good from code side but I get a weird error trying to open packs on this one, either mismatched # of renders or a unwrapped Text error

PR #723, opened 2024-02-28
Resolve frontend integration problems for trip related routes

sheharyarahmad842:
Some issues that are still to be resolved:

- We have to create 2 packs to create a relationship with a trip. When you create a single pack, it doesn't recognize that.
- Map is not getting displayed for larger cities.
- Page must be refreshed when we edit or delete a trip in the "Trip Create" or "Trip Detail" page to see the results.
- Item modal fields are not cleared after creating a new item.
- The search bar keeps showing options even if one is selected.


server/src/controllers/getOsm/getTrailsOSM.ts
```diff
@@ -29,8 +30,9 @@ import * as validators from '@packrat/validations';
 // };
 
 export function getTrailsOSMRoute() {
-  return publicProcedure.input(validators.getTrailsOSM).query(async (opts) => {
-    const { lat = 45.5231, lon = -122.6765, radius = 50000 } = opts.input;
+  return publicProcedure.query(async (opts) => {
+    // const { lat = 45.5231, lon = -122.6765, radius = 50000 } = opts.input;
+    const { lat = 45.5231, lon = -122.6765, radius = 50000 } = opts.rawInput;
```
andrew-bierman, line 35: just curious why all the rawInput instead of standard trpc input ?

server/src/controllers/getOsm/getTrailsOSM.ts
```diff
@@ -29,8 +30,9 @@ import * as validators from '@packrat/validations';
 // };
 
 export function getTrailsOSMRoute() {
-  return publicProcedure.input(validators.getTrailsOSM).query(async (opts) => {
```
andrew-bierman, line 32: why removing the validation ?

PR #729, opened 2024-02-29
native-base import to packrat/ui

pinocchio-life-like:
Change library import from native-base to own ui library


andrew-bierman:
We don't need to alias these just use the standard names from ui folder and then this is good.

Are there any remaining native base usages ? We should replace if so

andrew-bierman (approved):

andrew-bierman:
nice job, we still have a couple lingering native base components to swap out though based on file search

PR #743, opened 2024-03-07
Type fix on server is complete

pinocchio-life-like:
Type fix on server is complete, now when you run yarn check-types, there is no error on server.


andrew-bierman (approved):
nice job

PR #758, opened 2024-03-16
Type fix for Screens Directory

pinocchio-life-like:


pinocchio-life-like:
Still working on it, not done!

andrew-bierman:
Are all of these truly optional (as in do they have defaults that we use instead) ?

pinocchio-life-like:
> Are all of these truly optional (as in do they have defaults that we use instead) ?

I don't believe so, but When component is called or loaded these props are required but don't have value, so they are throwing type error, or we have passed some of the props but some are empty, so typescript is inferring type error. 

I am not exactly sure if they are optional but if we are passing some properties and leaving some when we call the components in other places, I assumed they might be optional.

andrew-bierman:
Sorry for being late on this one, but I went back through and think these changes should be good to merge. Let's rebase or copy to clean branch and I will approve. @jeweltee please do brief review once this gets updated

jeweltee:
@andrew-bierman sure, will do.

PR #760, opened 2024-03-17
Fix/expo app fix

taronaleksanian:


andrew-bierman:
Are you able to get submissions to pass validations? For some reason like adding a pack / trip is failing validations on my end. Maybe need to make a form schema with less constraints and only true inputs and attach the rest in handle submit function ?

taronaleksanian:
> Are you able to get submissions to pass validations? For some reason like adding a pack / trip is failing validations on my end. Maybe need to make a form schema with less constraints and only true inputs and attach the rest in handle submit function ?

Yeah we had an issue with forms and I fixed it.

andrew-bierman (approved):
Lgtm

PR #779, opened 2024-03-22
Muhammad hassan03 andrew testing

andrew-bierman:
@EngrAJ8888 should display preview here


packages/app/components/item/ItemForm.tsx
```diff
@@ -107,6 +109,12 @@ export const ItemForm = ({
               </RText>
             </SubmitButton>
           )}
+          <FormInput
```
andrew-bierman, line 112: does this impact adding items globally ie on /items route ?

PR #791, opened 2024-03-28
give max-width style to dialog content for native

Isthisanmol:
fixes add item modal ui on native
before
after


andrew-bierman:
Merging this in to fix broken ui but this should probably be a bottom sheet since on sm screen?

Isthisanmol:
> Merging this in to fix broken ui but this should probably be a bottom sheet since on sm screen?

I think it should not be a bottom sheet as bottom sheets are  generally used for supplementary content and actions.
I try to follow google's [guidelines](https://m3.material.io/components/bottom-sheets/guidelines).

andrew-bierman (approved):

PR #797, opened 2024-03-31
Fix pack  items table update on edit

Isthisanmol:
Fixes edited item update in pack items table


taronaleksanian:
@Isthisanmol can you please check if you see the same issue you have fixed by this PR [here](https://github.com/andrew-bierman/PackRat/pull/760). Your solution is a good option, but it seems we update the tanstack-query cache once item created/updated (see useEditPackItem hook)

PR #803, opened 2024-04-03
⬆️ expo 50 upgrade 

andrew-bierman:


andrew-bierman:
github actions failing likely due to next + expo adapter. Do we still need this ?

PR #805, opened 2024-04-03
update nextjs version to 14

Haysam-Bin-Tahir:


andrew-bierman:
export logic needs to be updated i think that is cause of failing actions

andrew-bierman:
Expo next adapter might need to be removed in favor of the tamagui one. I think that's causing issues

andrew-bierman (changes requested):
see github actions failing

PR #813, opened 2024-04-04
Setup `vitest` and add test cases to server

awais-codes:


andrew-bierman (approved):
nice job getting this set up :)

PR #817, opened 2024-04-05
UI update issues on mobile web

Tadjaur:
- [x] Fix profile view
- [X] Fix pack details view
- [X] Fix item option dropdown
- [X] Fix items list view

### DIFF
--- a/packages/ui/src/ZDropdown/index.tsx
+++ b/packages/ui/src/ZDropdown/index.tsx
@@ -68,13 +68,13 @@ const CustomContent = styled(DropdownMenu.Content, {
             style={{ padding: 0}}
           />
         </DropdownMenu.Trigger>
-        <DropdownMenu.Content>
+        <CustomContent>
           {dropdownItems.map(({label, onSelect = () => {}}) => (
             <DropdownMenu.Item key={label} onSelect={onSelect()}>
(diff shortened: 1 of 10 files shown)


andrew-bierman (approved):
nice job

packages/ui/src/ZDropdown/index.tsx
```diff
@@ -68,18 +68,18 @@ const CustomContent = styled(DropdownMenu.Content, {
             style={{ padding: 0}}
           />
         </DropdownMenu.Trigger>
-        <DropdownMenu.Content>
+        <CustomContent>
```
andrew-bierman, line 71: this is only change i'm not sure about, why do we need this ?

Tadjaur: The default `DropdownMenu.Content` is transparent sir, And the `CustomContent` defined 50 lines above fix that and it also add a custom style.
Here is the code 
```ts
const CustomContent = styled(DropdownMenu.Content, {
        backgroundColor: 'white',
        minWidth: 160,
        shadowColor: '#000',
        borderRadius: 8,
        shadowOffset: {
            width: 0,
            height: 8,
        },
        shadowOpacity: 0.2,
        shadowRadius: 16,
        padding: 12,
    });
  ```

PR #824, opened 2024-04-06
Fix/fix validations

taronaleksanian:


packages/app/components/chat/index.tsx
```diff
@@ -14,12 +14,14 @@ import {
...
   SubmitButton,
 } from '@packrat/ui';
-// import { sendMessage } from '@packrat/validations';
+import { sendMessage } from '@packrat/validations';
 import useCustomStyles from 'app/hooks/useCustomStyles';
 import { useChat } from 'app/hooks/chat/useChat';
 import { loadStyles } from './chat.style';
 // import { Select } from "tamagui";
 
+// TODO check if we fixed chat screen on another branch
```
Tadjaur, line 23: @taronaleksanian I suggest you create an issue on the github repository and paste the link of this created issue here.

Reasons: 
- It helps by providing more context about the task
- We can easily keep track of the tasks that need to be done.

```suggestion
// TODO check if we've fixed the chat screen on another branch
// link: https://github.com/andrew-bierman/PackRat/issues/ ???
```

packages/app/components/password-reset/PasswordResetAlert.tsx
```diff
@@ -0,0 +1,42 @@
...
+    <YStack
+      style={{
+        backgroundColor: 'rgba(52, 168, 154, 0.7)',
+        padding: 24,
+        borderRadius: 8,
+        textAlign: 'center',
+      }}
+    >
+      <RText style={{ color: 'white', fontWeight: 'bold', fontSize: 24 }}>
+        Password reset successfully!
```
Tadjaur, line 17:
```suggestion
        Password reset completed successfully!
```
or 

```suggestion
        password reset succeeded!
```

packages/app/hooks/user/useUpdateUser.ts
```diff
@@ -1,18 +1,16 @@
...
+  const { mutateAsync: mutateUserAsync } = queryTrpc.editUser.useMutation();
   const setUser = useUserSetter();
 
   const updateUser = (user) => {
     (async () => {
       try {
-        const updatedUser = await mutateUserAsync(user);
+        const { userId, ...rest } = user;
+        const updatedUser = await mutateUserAsync({ id: userId, ...rest });
+        // @ts-ignore
```
Tadjaur, line 13: We should always avoid disabling linter in code.

I recommend that you fix the linting problem or add more comments on why you needed to disable lint here.

All it takes is for one developer to start disabling linting (perhaps for a good reason) for other developers to start doing the same thing everywhere, but for lazy reasons :) 

``suggestion
```

File: packages/validations/src/validations/tripRoutesValidator.ts [hunk 4]
```diff
@@ -1,52 +1,80 @@
...
 
-export const JoiObjectId = (message = 'valid id') =>
-  z.string().regex(/^[0-9a-fA-F]{24}$/g, { message });
+export const addTripForm = z.object({
+  name: z.string().nonempty(),
+  description: z.string().nonempty(),
+  isPublic: z.union([z.literal('0'), z.literal('1')]),
+});
+
+// @ts-ignore
```
Tadjaur, line 9: Same comment as above.
```suggestion
```

packages/validations/src/validations/userRoutesValidator.ts
```diff
@@ -1,81 +1,100 @@
...
-const JoiObjectId = z.string().regex(/^[0-9a-fA-F]{24}$/g);
+const emailValidator = z
+  .string()
+  .email()
+  .transform((str) => str.trim().toLowerCase());
+
+const passwordValidator = z
+  .string()
+  .min(7)
+  .refine((str) => !str.includes('password'), {
```
Tadjaur, line 11: Is there any reason why we shouldn't allow the password to contain the word "password"?

If not, can we instead check whether the password contains at least one uppercase letter, at least one number, at least one non-alphanumeric character and should not match the username?

andrew-bierman (approved):
Lgtm

PR #825, opened 2024-04-08
Feat/vite static site

andrew-bierman:


andrew-bierman:
great solution for reference we could implement. need to adapt to work with expo icons.  https://github.com/oblador/react-native-vector-icons/issues/1477

Tadjaur (approved):

PR #832, opened 2024-04-11
Add EAS android preview  build workflow

Isthisanmol:
-loaded environment varibles as secret on EAS
-remove gradle.properties and app.json from git ignore


andrew-bierman (approved):
Great job :)

PR #880, opened 2024-05-01
Fix: Owner View in Trips Page

MuhammadHassan03:
In the trips page, click on view owner redirects to the owner profile


packages/app/components/pack_table/Table.tsx
andrew-bierman, line 1: It should be zeego menu for all crud not individual columns fyi see how we handled pack items

MuhammadHassan03: noted that, i'll change it rn..

MuhammadHassan03: I have updated the table items with zeego menu implementation in trips creation page please check it out.

andrew-bierman (changes requested):
Looks good. Thanks 

Just clean this PR up when possible please, what changes are actually needed to fix this issue? 

- hardcoded keys 
- logs
- migration changes

MuhammadHassan03:
@andrew-bierman I have tried many times to clean this PR but as u have elobrated to me that remove all console.log's, i did same but still it shows conflict in these files but i cannot see any conflict in my local branch...:(

PR #898, opened 2024-05-05
Fix eas cloud build part-1

Tadjaur:


packages/app/envs/index.native.ts
```diff
@@ -0,0 +1,14 @@
+import Constants from 'expo-constants';
```
andrew-bierman, line 1: can we handle this with the packages/config that we have ?

Tadjaur: We should then have to create the `.env` file with all variable prefixed by `EXPO_PUBLIC_`.

Tadjaur: Here is the list of update I plan to do:
- [ ] update imports from `@env` or `app/env` to `@packrat/config`
- [ ] update the workflow `android-preview-build-local.yml` to defined these variables
- [ ] update the readme

andrew-bierman: @Tadjaur see the recent changes to the github action, can we map over these to set the env variables or will we have to call export manually for each one?

packages/app/components/map/WebMap.web.tsx
```diff
@@ -1,15 +1,12 @@
 import React, { useState } from 'react';
 import mapboxgl from 'mapbox-gl';
-import { MAPBOX_ACCESS_TOKEN } from '@env';
```
andrew-bierman, line 3: we should use this instead of hardcoded key below, did not realize that was in there lol

Issue #843, opened 2024-04-16
Mobile Web Issues 


JewelTee:
- [x] Search item field not available on mobile web 
![438083196_1594507341324824_2494821132267699606_n](https://github.com/andrew-bierman/PackRat/assets/166146993/f6c5a17a-b482-4645-94af-682e0f8c0d3e)

- [x] Appearance and Logout option missing on mobile web
![438060632_408687995128921_4542354910636225110_n](https://github.com/andrew-bierman/PackRat/assets/166146993/7bd54de2-2b0b-4f94-aa70-8b31acab4662)

EngrAJ8888:
Adding the following issues on this ticket:

In the mobile web version

- [ ] Users are unable to mark packs as favorites directly from the feed page, as the heart button is missing. Currently, this action can only be performed using the web version on a PC. 
![43](https://github.com/andrew-bierman/PackRat/assets/122357250/6254322a-780c-4b02-b01c-bfa173e1fca8)

- [ ] Update the display and alignment of text in the Pack and scoring section. 
![42](https://github.com/andrew-bierman/PackRat/assets/122357250/1e8d8358-a824-40ee-a3fd-07f4c37c108f)

Isthisanmol:
> * [ ]  Search item field not available on mobile web
[...]

PR #912, opened 2024-05-08
[#843] Mobile web issues fix

mikeCyb:
**Fixes**
1. The heart button in packs page is fixed.
2. Scoring section text alignment improved.

### DIFF
--- a/packages/app/components/feed/FeedCard.tsx
+++ b/packages/app/components/feed/FeedCard.tsx
@@ -281,7 +281,7 @@ export default function Card({
                             gap: 8,
                           }}
                         >
-                          {user?.id === owner_id ? null : (
+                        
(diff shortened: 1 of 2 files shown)


andrew-bierman:
Can you please submit this PR via clone so we get a preview link for @jeweltee ?

packages/app/components/feed/FeedCard.tsx
```diff
@@ -281,7 +281,7 @@ export default function Card({
                             gap: 8,
                           }}
                         >
-                          {user?.id === owner_id ? null : (
```
andrew-bierman, line 284: A user should not be able to favorite their own stuff, do we need to remove this ?

mikeCyb: Okay, thought the other way around. I will be fixed.

PR #917, opened 2024-05-09
Trip improvments

taronaleksanian:


EngrAJ8888:
Hi @taronaleksanian - The update looks great! I'd like to add that trip details should be accurately reflected when viewing trips. I created a trip with Biking and Running as activities, but upon checking, the Activity section only displays the word "Trip" for both. Could you please take a look? I've attached a photo for reference. Thank you! 

image
image

taronaleksanian:
Sure I will check it thanks

andrew-bierman:
Good stuff, this would be cool if it filtered the search bar and nearby cards based on activity type ie selecting biking should not show a standard footpath. 

I believe we can update osm queries to optionally accept this new type

taronaleksanian:
Yeah, great idea will implement it too!

PR #918, opened 2024-05-09
Make CopyPackModal component in packs

Isthisanmol:
Make Copy pack modal component in packs

### DIFF
--- a/packages/app/components/card/CustomCardHeader.tsx
+++ b/packages/app/components/card/CustomCardHeader.tsx
@@ -21,7 +21,7 @@ export const CustomCardHeader = ({ data, title, link, actionsComponent }) => {
       }
       </RStack>
       <View>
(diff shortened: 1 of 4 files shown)


jeweltee:
@Isthisanmol hi! Thanks for this update. The Copy Pack feature is working fine, but the page loads a bit slowly after copying a pack, and the text alignments are broken. The edit and delete options are not working as well. Please check the screen recording below for reference. Thank You! 


https://github.com/andrew-bierman/PackRat/assets/166146993/0e44265b-60a1-4113-bd34-1755f986cab6

Screenshot 2024-05-12 at 20 41 19

EngrAJ8888:
Hi @Isthisanmol, I've noticed that the copy option is currently available only when users open their own packs. However, it should be visible when users open packs created by others. The idea is that when users browse packs from other users in the global, they should have the option to copy all items from those packs. I hope that helps. Thanks!

Issue #883, opened 2024-05-01
Trip Creation Functionalities 


JewelTee:
- [ ] The response of showing the Trip Weather, Map, Nearby Trails, and Nearby Parks is a little slow upon choosing a destination

- [ ] Weather and map data are not showing up on some destinations

MuhammadHassan03:
I have started working on the second one issue

jeweltee:
- [ ] Make the "Save" button more visible 
<img width="1440" alt="Screenshot 2024-05-04 at 12 36 56" src="https://github.com/andrew-bierman/PackRat/assets/166146993/55f046e8-f45a-4fcd-8d64-fe7159e9249c">

jeweltee:
> I have started working on the second one issue

hi. Can you also include fixing the map from the destination query page, maps are empty on other destinations as well. Thanks!

MuhammadHassan03:
> > I have started working on the second one issue
> 
> hi. Can you also include fixing the map from the destination query page, maps are empty on other destinations as well. Thanks!

Yeah, am working on it

jeweltee:
- [ ] Arrows in the Create Trip page are not working. Please see the screen recording below for reference. Thank You! 

https://github.com/andrew-bierman/PackRat/assets/166146993/bda849d7-48de-46a4-b9b4-2f1d60c0b44a

jeweltee:
- [ ] When you navigate to the trips page and click on "view owner" it shows an error. Please refer to the image below 
<img width="1436" alt="Screenshot 2024-05-08 at 23 02 34" src="https://github.com/andrew-bierman/PackRat/assets/166146993/a0ced7f0-e801-4d19-98e4-885a5873584b">

andrew-bierman:
- [ ]  Full screen map is broken on web

awais-codes:
> * [ ]   Full screen map is broken on web

I will be working on this one

Issue #841, opened 2024-04-15
Addressing Existing System Issues/Bad UI and Prioritizing Fixes


EngrAJ8888:
- [x] The 'Next page' button on the global items page is not working. Users can view additional items beyond the first page only by selecting a value greater than 10 from the dropdown menu below.
- [x] Remove edit/ delete column in global items
![image](https://github.com/andrew-bierman/PackRat/assets/122357250/18cc9309-d913-41f6-b2f7-3ea6d023e5b1)

- [x] Fix the alignment of pack details, this issue is visible when viewing other users' packs in the feed.
![40](https://github.com/andrew-bierman/PackRat/assets/122357250/a3b16e74-4183-49ba-b073-e6102ab71a99)

- [x]  Update the search field on the "create a pack" page to be centered.
![41](https://github.com/andrew-bierman/PackRat/assets/122357250/b3b0d8f5-4250-4bfa-b4fe-b90bd9b06cda)

jeweltee:
- [ ] Resize "Add Pack" button (It's too wide) 
![screenshot-1713191216431](https://github.com/andrew-bierman/PackRat/assets/166146993/54d12c67-e00f-4672-a128-bf0693998abb)

jeweltee:
- [ ] search menu dropdown is overlapping on other text field
![screenshot-1713192428408](https://github.com/andrew-bierman/PackRat/assets/166146993/d4333462-716c-4903-9ad6-f90a40c5e168)

EngrAJ8888:
- [ ] In the mobile web version, update the display of the Pack and scoring section.
![42](https://github.com/andrew-bierman/PackRat/assets/122357250/1e8d8358-a824-40ee-a3fd-07f4c37c108f)

andrew-bierman:
Let's make sub issues for these to track

EngrAJ8888:
Hey @andrew-bierman, the ticket has been updated accordingly.

@shahzi113awanm, thank you handling this task. We've added the latest list of required updates to the ticket. If you have any questions or need clarification on anything, don't hesitate to reach out to me or @jeweltee. 

Thank you!

shahzi113awan:
<img width="516" alt="Screenshot 2024-04-18 at 8 29 03 AM" src="https://github.com/andrew-bierman/PackRat/assets/34761350/f0729783-2240-4c5a-a9c3-30694de034a8">

shahzi113awan:
having some zod errors while creating pack 
[...]

PR #928, opened 2024-05-12
[#841] Bad UI fixes and improvements 

mikeCyb:
**fixes**
1. Edit/ delete column is removed in global items.
2. Search menu dropdown overlapping is now fixed.

Additionally it has Trip creation functionality fixes[#883], But I have asked PR last week for them.


jeweltee:
@mikeCyb hi! thanks for this update. Regarding the search menu dropdown, I noticed that when searching items in global, the form for items moves down to prevent overlap. Can we update this so that the form containing the items doesn't move downward? We could do it like the search menu dropdown on the Create Trips page or something similar. Thank You! 

The Screenshot below is the search menu dropdown on the Create Trips Page
aa0e4e1b-004f-43d1-96df-eae1fa5db418

mikeCyb:
@jeweltee : Thank you for the clarification; it will be addressed.

jeweltee:
@mikeCyb hi! since you opened a new ticket addressing the issue on the search menu dropdown (#947), can we update this PR and remove the update on the Search menu dropdown? so this will only be for the 'Edit/ delete column is removed in global items.' To avoid conflicts.  Thanks!

jeweltee:
@mikeCyb hi! Just want to follow up on this. Thanks!

jeweltee:
@mikeCyb hi! Just want to follow up on this. Thanks!

PR #931, opened 2024-05-13
Fix/text formatting

pinocchio-life-like:


packages/app/components/card/CustomCardHeader.tsx
```diff
@@ -21,7 +21,7 @@ export const CustomCardHeader = ({ data, title, link, actionsComponent }) => {
       }
       </RStack>
       <View>
-        <Link href={`/profile/${data.owner_id.id || data.owner_id}`}>
+        <Link href={`/profile/${data.owner_id.id || data.owner_id}`} style={{ textDecoration: 'none' }}>
```
andrew-bierman, line 24: We should probably add this to the Link we export

pinocchio-life-like: > We should probably add this to the Link we export

@andrew-bierman Yes, but the Link is getting imported from crosspath as a function declaration not a component, and is even in .d.ts file.

So Instead I have defined the same Link component inside ui directory and Imported the Link from crosspath file and set the style property on it now, and imported that for usage. I hope that is okay. lmk please!

andrew-bierman: Got it, yes that is a good solution. We should treat link as a UI component I believe, considering it does apply styles

jeweltee:
@pinocchio-life-like the update looks good! Thank you

andrew-bierman (approved):

jeweltee (commented):
Update looks good

PR #936, opened 2024-05-15
Fix config for android local build

Tadjaur:


.github/workflows/expo.build.yml
```diff
@@ -10,6 +10,12 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+env: # Define a shared environment variable for android and ios build.
+  MAPBOX_DOWNLOADS_TOKEN: ${{ secrets.MAPBOX_DOWNLOADS_TOKEN }}
+  EXPO_PUBLIC_CLIENT_URL: ${{ vars.EXPO_PUBLIC_CLIENT_URL }}
+  EXPO_PUBLIC_API_URL: ${{ vars.EXPO_PUBLIC_API_URL }}
+  EXPO_PUBLIC_MAPBOX_ACCESS_TOKEN: ${{ secrets.EXPO_PUBLIC_MAPBOX_ACCESS_TOKEN }}
```
Tadjaur, line 17: @andrew-bierman  These variable must be defined. 
here I'm using some `vars` keyword but if these variables are all defined as secrets just let me know and I will update them.

apps/expo/env.example
```diff
@@ -1,23 +1,9 @@
...
-GOOGLE_PLACES_API_KEY = 'Your api key'
-MAPBOX_API_KEY = "Your api key, starts with pk..."
-OPENWEATHER_KEY = "Your api key"
-X_RAPIDAPI_KEY = "Your api key"
-MAPBOX_ACCESS_TOKEN = "Your api key, starts with pk..."
+EXPO_PUBLIC_CLIENT_URL="Your client url"
+EXPO_PUBLIC_API_URL="Your api url"
+EXPO_PUBLIC_MAPBOX_ACCESS_TOKEN="Your api key, starts with pk..."
+EXPO_PUBLIC_ANDROID_CLIENT_ID="962352557394-........................apps.googleusercontent.com"
+EXPO_PUBLIC_IOS_CLIENT_ID="962352557394-........................apps.googleusercontent.com"
```
Tadjaur, line 7: Only environment variable in use is defined here.

.github/workflows/expo.build.yml
andrew-bierman: I will define these as secrets just to keep everything in the same place, github secrets will be updated with these values momentarily

Tadjaur: @andrew-bierman updated.

andrew-bierman (approved):
nice job :)

PR #951, opened 2024-05-18
fix modal not showing done

pinocchio-life-like:


jeweltee:
@pinocchio-life-like the full-screen map is now fixed. Thanks!

@andrew-bierman - this is good to go for merging

andrew-bierman (approved):

PR #975, opened 2024-06-01
Enhance/chat

pinocchio-life-like:


EngrAJ8888:
Hi @pinocchio-life-like , Thanks for the update! Everything looks great. However, I'd like to suggest one addition. When I tried minimizing the chat (by clicking the chat icon only), I noticed that the text I entered disappeared. Could we update this so that the chat history is retained? Perhaps we could have the "X" button on top clear the chat history instead.
image

pinocchio-life-like:
Hi @EngrAJ8888 which text do you mean exactly? the one on the input field or those you have already sent?

if its the one you have already sent, since the backend is returning error, sent messages will not be retained. but if you mean the text field being cleared, I will address that. let me know!

EngrAJ8888:
Hi @pinocchio-life-like , I was referring to the text I inputted into the chat AI. I've also attached a photo for reference.
image

Does this mean we'll leave it as is for now, considering it's a backend issue? Thanks!

pinocchio-life-like:
@EngrAJ8888 Yes, I believe.

Screenshot 2024-06-03 140118

I have left message for on discord. check that.

EngrAJ8888:
Hi @pinocchio-life-like - Please refer to the video for clarification on the issue I raised. The text should remain if I only minimize the chat box. However, if the problem lies in the backend, we can go ahead and close this out. Thank you! 

https://github.com/andrew-bierman/PackRat/assets/122357250/1b57be61-de9d-4ac3-b402-1dc91e4c5d64

pinocchio-life-like:
@EngrAJ8888 Yes I am sure that this is OpenAI API key not being set for the backend. If this suffices the issue you raised you may close this. Its also okay if you dont since I am working on other enhancements regarding chat.

pinocchio-life-like:
Update, it works now.

EngrAJ8888:
Hi @pinocchio-life-like - thanks for the update! It's looking great!

@andrew-bierman - this is good for merging.

packages/app/components/pack/PackDetails.tsx
```diff
@@ -63,13 +70,13 @@ export function PackDetails() {
...
-            ? { minHeight: '100vh' }
-            : { minHeight: Dimensions.get('screen').height },
+          // Platform.OS == 'web'
+          //   ? { minHeight: '100vh' }
+          //   : { minHeight: Dimensions.get('screen').height },
         ]}
       >
-        {!isError && (
-          <>
+        <ScrollView showsVerticalScrollIndicator={false}>
```
andrew-bierman, line 78: this causes weird double scroll effect on web, we should avoiding nesting them

pinocchio-life-like: Yes but this was meant to make the chat opener float inside the view. or may  be I will use fixed for web instead.

PR #979, opened 2024-06-06
🐛 add env values for tests

andrew-bierman:


andrew-bierman:
Still a few failing. Merging but will need to circle back to this

PR #983, opened 2024-06-08
Change width according to screen size

Isthisanmol:


taronaleksanian (approved):
Great solution, thanks. Only we can do one improvement.  I think it would be better if we store "425" into a variable. app/constants/breakpoints.ts can be a good place to store that variable

PR #1000, opened 2024-06-12
Sort name and alignment changes

Isthisanmol:


andrew-bierman (approved):
Lgtm, note tamagui has a built in way we can handle these media queries so we do not need a much Platform.OS logic

JewelTee:
@Isthisanmol the update looks good! 👍🏻 Thank You

Isthisanmol:
> Lgtm, note tamagui has a built in way we can handle these media queries so we do not need a much Platform.OS logic

I'll replace that and keep that in mind for future.

PR #1021, opened 2024-06-18
replace sheet to zeego dropdown in RSelect

Isthisanmol:


packages/ui/src/RSelect/index.tsx
```diff
@@ -64,83 +54,62 @@ export function SelectItem(props) {
...
-  }
-
   return (
     <Select
       value={value}
       disablePreventBodyScroll
       onValueChange={handleChange}
       {...forwardedProps}
     >
-      <Select.Trigger width={220} iconAfter={ChevronDown}>
```
andrew-bierman, line 97: Nice job but I think we should support both via prop is my thinking

just select with prop of native = sheet
Select with prop native + zeego = zeego dropdown

Isthisanmol: @andrew-bierman Could you please review the code? If it is correct, I will apply it to the other instances where we use dropdown menus.

andrew-bierman: yes this looks good, nice job :)

JewelTee:
@Isthisanmol the dropdown looks great! Can we also change the dropdown from the weight unit when adding new item? So it won't occupy most of the screen. Thank You! Please see image below: 
Screenshot_20240621_143056

JewelTee:
@Isthisanmol everything looks good. Thank You!

@andrew-bierman - this is ready for merging 👍🏻

packages/app/components/Dropdown.tsx
```diff
@@ -116,34 +116,45 @@
```
andrew-bierman, line 116: we should probs move this into ui package, not sure why its in components (not your fault, just noticed this)

andrew-bierman (approved):

packages/app/components/Dropdown.tsx
Isthisanmol: Done that.

PR #1031, opened 2024-06-21
theme state on reload fix

pinocchio-life-like:


packages/app/context/theme.tsx
```diff
@@ -59,6 +61,28 @@ const ThemeContext = createContext({
 export const ThemeProvider = ({ children }) => {
   const [state, dispatch] = useReducer(reducer, initialState);
 
+  useEffect(() => {
+    const fetchTheme = async () => {
+      let storedIsEnabled;
+      if (Platform.OS === 'web') {
```
andrew-bierman, line 67: pretty sure this will crash on native, but we have an abstracted hook for accessing storage on either native/web that we could use here

JewelTee (commented):
@pinocchio-life-like the issue is still persisting. Please check the video below. Thanks!

https://github.com/andrew-bierman/PackRat/assets/166146993/e1ceef1a-3a3b-407a-8b4d-cd50841329f6

pinocchio-life-like:
@JewelTee I don't think its the theme that is flashing, that is a different issue actually, you can check it with light mode as well, the page that is flashing is actually a landing page that you see before you login. that is because we are checking whether the user is logged and deciding which landing page to load, either the authorized or the unauthorized. you can almost see it clearly in both light and dark mode before it loads the correct page.

JewelTee:
@pinocchio-life-like oh okay. Got it! Thank You.

@andrew-bierman - this is ready for merging 👍🏻

packages/app/context/theme.tsx
pinocchio-life-like: Ok, I did change it to use the hook now. check it please

pinocchio-life-like:
@andrew-bierman I have made it to use the abstracted storage hook but the theme flashes, why not use localstorage for web in our hook?

packages/app/context/theme.tsx
```diff
@@ -1,12 +1,16 @@
-import { createContext, useReducer } from 'react';
+import { createContext, useEffect, useReducer, useState } from 'react';
 import { theme, darkTheme } from '../theme';
 import ThirdPartyThemeProviders from './ThirdPartyThemeProviders';
 import React from 'react';
+import { Platform } from 'react-native';
+import AsyncStorage from '@react-native-async-storage/async-storage';
```
taronaleksanian, line 6: We have storage util to save data on the storage, can't we  just use that for storing data?

packages/app/hooks/appearance/useAppearance.ts
```diff
@@ -0,0 +1,28 @@
...
+    setIsEnabled(newIsEnabledValue);
+    if (!isEnabled) {
+      enableDarkMode();
+    } else {
+      enableLightMode();
+    }
+  };
+
+  return {
+    isEnabled,
```
taronaleksanian, line 22: We don't use this hook anymore, we can delete it

pinocchio-life-like: I am not sure if we should delete it, cause its meant to separate concern out of appearance page on toggle switch.

packages/app/context/theme.tsx
pinocchio-life-like: Ok, I will use that. that definitely seems a better option.

andrew-bierman (approved):

PR #1032, opened 2024-06-22
Fix app crash when viewing downloaded maps

Tadjaur:


EngrAJ8888:
Hi @Tadjaur , I can now view the saved maps without any errors. However, I am unable to scroll down the screen. Could we update this? Once done, we're good to merge. Thanks!

JewelTee (commented):
I was able to download and view downloaded maps as well but when I tried to download for the second time the app crashes.

andrew-bierman (approved):
nice job :)

PR #1038, opened 2024-06-25
Feat/UI library

taronaleksanian:


packages/app/constants/api.ts
```diff
@@ -6,4 +6,4 @@ import { API_URL } from '@packrat/config';
  * format: "{scheme}://{serverhost}:{port}/api"
  * e.g: "http://localhost:4200/api"
  */
-export const api = API_URL;
```
andrew-bierman, line 9: Revert

andrew-bierman (approved):
Lgtm

packages/app/constants/api.ts
taronaleksanian: @andrew-bierman fixed

PR #1043, opened 2024-06-26
feed styling

mnandwere:
Modified Feed filter styles.
Search text clearing issue on filter.


packages/app/components/feed/SearchProvider.tsx
```diff
@@ -0,0 +1,16 @@
+import React, { createContext, useState } from 'react';
+
+// Create a context with default value
+const SearchContext = createContext();
```
Isthisanmol, line 4: The search is working  on native but the useContext is undefined on web. I think createContext should be provided some default value to avoid this.

packages/ui/src/form/components/useForwardFocus.tsx
```diff
@@ -0,0 +1,23 @@
...
+  return {
+    onFocus: () => {
+      setFocused(true);
+    },
+    onBlur: () => {
+      setFocused(false);
+    },
+    focusable: true,
+  };
+};
```
Isthisanmol, line 23: I think this hook should be moved to packages/app/hooks.

packages/app/components/feed/SearchProvider.tsx
andrew-bierman: Wrap the /screen component in context not the expo router should fix this

EngrAJ8888:
Hi @mnandwere, the updates on the web are functioning as expected. However, I've encountered an issue with the sorting function in the native app—it's not working accurately. Thank you!

JewelTee (commented):
@mnandwere I encountered the same issue @EngrAJ8888 has mentioned

EngrAJ8888:
Hi @mnandwere , the search text clearing is now functional. However, I'm still facing the same issues as before. When I select the sorting option "Most Items," it displays items with the least number of items on both feeds and packs screens. Could you please look into this?
Additionally, could you update this to the latest Andrew testing branch? The trips pages should be hidden for now. Thank you!

EngrAJ8888 (approved):

PR #1053, opened 2024-06-30
Feat/delete account

taronaleksanian:


taronaleksanian:
@JewelTee @EngrAJ8888 
You can check this for the users with no Google auth, I'll check if it works for Google auth and let you know, maybe there is additional work needed for the Google auth users, but workflow stays the same for both

andrew-bierman (approved):
Looks good to me from quick read

taronaleksanian:
@JewelTee @EngrAJ8888  I checked it for the Google auth users and it works, so you can check it now

PR #1058, opened 2024-07-02
fix profile update native

pinocchio-life-like:


EngrAJ8888:
Hi @pinocchio-life-like , thank you for the update! I can now update details from the user profile settings. However, when I return to the profile, it shows null (please refer to the photo below)
image

PR #1068, opened 2024-07-04
Feat/remaining types errors

MuhammadHassan03:


packages/app/components/user/UserDataCard.tsx
```diff
@@ -147,7 +147,7 @@ const UserDataCard = ({
                   // marginTop: '-3px',
                 }}
               >
-                Total Weight: {total_weight.toFixed(2)}g
+                Total Weight: {total_weight?.toFixed(2)}
```
pinocchio-life-like, line 150: I think we need a unit that represents the weight here. it can be gram or may be we can get it from user profile and set the preferred unit. either way

packages/app/hooks/chat/useChat.ts
```diff
@@ -34,10 +34,7 @@ export const useChat = (itemTypeId = null) => {
 
   const [userInput, setUserInput] = useState('');
 
-  const { data: chatsData, refetch } = useGetUserChats(
-    user.id,
-    typeId.itemTypeId,
-  );
+  const { data: chatsData, refetch } = useGetUserChats(user.id, typeId);
```
pinocchio-life-like, line 37: typeId is an object with a property itemTypeId, so this will fail to fetch the user chat I believe.

packages/app/screens/user/ProfileContainer.tsx
```diff
@@ -198,6 +198,17 @@ export default function ProfileContainer({ id = null }) {
     error,
   } = useProfile(id);
 
+  useEffect(() => {
```
pinocchio-life-like, line 201: clean this

pinocchio-life-like (approved):
@MuhammadHassan03 Overall, this looks really nice, though I have added some cleanups.

packages/app/hooks/chat/useChat.ts
pinocchio-life-like: You can check this pack: https://packrat.pages.dev/pack/jfxuhkd3zflt4ddavnyhh3so from testing preview, it has user chats but your PR preview: https://73a8f0cc.packrat.pages.dev/pack/jfxuhkd3zflt4ddavnyhh3so for the same pack doesn't. That might be why

EngrAJ8888:
Hi @MuhammadHassan03, I can now create packs, but there's no option to search for global items. Please refer to the photo below.
image

packages/app/screens/user/ProfileContainer.tsx
MuhammadHassan03: cleaned

packages/app/hooks/chat/useChat.ts
MuhammadHassan03: yeah you are right i forgot to add .itemTypeId after typeId

packages/app/components/user/UserDataCard.tsx
MuhammadHassan03: yeah i checked the unit not coming from anywhere in this so i, added the static g again here

EngrAJ8888:
Hi @MuhammadHassan03, looks great! Thanks!

@andrew-bierman - this is good to merge.

andrew-bierman (approved):

PR #1082, opened 2024-07-09
remove name attribute from stack component to fix native app crash

taronaleksanian:


andrew-bierman (approved):
Nice find, thanks. Merging this in

PR #1096, opened 2024-07-16
Feat/integrate rest api

pinocchio-life-like:


server/src/services/item/addItemService.ts
```diff
@@ -49,8 +53,8 @@ export const addItemService = async (
   });
 
   // await Pack.updateOne({ _id: packId }, { $addToSet: { items: newItem._id } });
-  await itemPacksClass.create({ itemId: item.id, packId });
```
andrew-bierman, line 52: why this change ?

andrew-bierman (approved):
nice job :) 

later on we should look at the honor swagger UI integration to replace the previous one

PR #1102, opened 2024-07-17
Fix chat icon visibility on native

Isthisanmol:


JewelTee (commented):
@Isthisanmol the chat icon is now visible on native, but can we maybe resize the chat box because it looks like this as of the moment. Thank You
Screenshot_2024-07-17-19-32-49-61_248012cd4184c9416d967f4c08ee0567

PR #1110, opened 2024-07-18
Fix double splash screen

Isthisanmol:


JewelTee (approved):
@Isthisanmol looks good to me 👍🏻

PR #1115, opened 2024-07-21
Server: save downloaded map in user profile

Tadjaur:
This PR enable on the server the ability to validate the downloaded map meta data and save  it inside the database.


andrew-bierman (approved):
Nice job :)

PR #1127, opened 2024-07-22
Import Items for packs and global items

pinocchio-life-like:


pinocchio-life-like:
Note: resolving some issue on this...

andrew-bierman:
Oh you did this all client side (not against it). is there anything you think we should move / copy over to server? would be nice to have some end points to hit that did this papa parse style logic so we can upload data from elsewhere

pinocchio-life-like:
@andrew-bierman Yes, but now I believe since we need to listen to endpoints, we can move this whole logic to the backend. => for the clients local file upload, we can upload the file to the bucket and stream read that on the backend and update the front end after its done. I think that will be better, to not make the client side too heavy.

pinocchio-life-like:
@andrew-bierman  What do you think?
Should I update this, or have like v2 of it?

packages/app/components/item/DocumentPicker/DocumentPickerNative.tsx
```diff
@@ -0,0 +1,2 @@
+import DocumentPicker from 'react-native-document-picker';
```
andrew-bierman, line 1: im fine with this but we might want to consider https://docs.expo.dev/versions/latest/sdk/document-picker/ just so we don't have to split web and native up

andrew-bierman:
> @andrew-bierman What do you think? Should I update this, or have like v2 of it?

Im good with this for now you accomplished the goal - but lets add some endpoints on server to accept csv for items and packs also

andrew-bierman (approved):
lgtm, consider the expo lib if its easy swap. usually their libraries work well for us

packages/app/components/item/DocumentPicker/DocumentPickerNative.tsx
pinocchio-life-like: Ok, good.

pinocchio-life-like:
@andrew-bierman  I have moved the logic to the backend, and made a global endpoint for it.

andrew-bierman (approved):

PR #1130, opened 2024-07-23
Feat/preview version of api

taronaleksanian:


.github/actions/setup-wrangler-toml/action.yml
andrew-bierman, line 1: Should we just use the env shim script we have but for backend it will generate .preview.vars and .vars for prod? Might be easier to maintain if we can get it functional. Then we could put the wrangler.toml on github if we can keep secrets out of it https://developers.cloudflare.com/workers/configuration/environment-variables/#add-environment-variables-via-wrangler

andrew-bierman (approved):
im good with these changes overall, i do think we might want to consider my comment about using the vars files so that we can keep wrangler.toml in source and not have to echo each value in

PR #1139, opened 2024-07-24
Fix validation hot reloading on native and server apps

Tadjaur:
This pr fix validation workspace hot reloading on server and expo application.

### DIFF
--- a/apps/expo/metro.config.js
+++ b/apps/expo/metro.config.js
@@ -15,7 +15,13 @@ config.transformer.minifierConfig.compress.drop_console = true;
 config.resolver.assetExts.push('cjs');
 
 // Include 'mjs' and 'cjs' in source extensions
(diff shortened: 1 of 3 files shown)


andrew-bierman:
Does adding the validations to the root ts config not work for this? 

I think server, UI and other packages all hot reload from that root ts config if I'm not mistaken. Ideally trying to keep it all together but not a big deal at all 

If root config does not work easily, these changes look fine, nice job :)

andrew-bierman (approved):
lgtm

Tadjaur:
> Does adding the validations to the root ts config not work for this?
> 
> I think server, UI and other packages all hot reload from that root ts config if I'm not mistaken. Ideally trying to keep it all together but not a big deal at all
> 
> If root config does not work easily, these changes look fine, nice job :)

I tried but it does not work.

PR #1156, opened 2024-08-05
Import from bucket for global items

pinocchio-life-like:


packages/validations/src/index.d.ts
andrew-bierman, line 1: looks like validations package might have been duplicated

andrew-bierman:
Good work here, gonna drop a few clean up comments but this is quite close

server/src/controllers/item/importFromBucket.ts
andrew-bierman, line 1: Let's pull this business logic into a service layer to clean it up

server/src/routes/index.ts
```diff
@@ -60,11 +35,55 @@ router.route('/favorite', favoriteRouters);
 router.route('/mapPreview', mapPreviewRouter);
 router.route('/health', healthRoutes);
 
-const helloRouter = new Hono();
-helloRouter.get('/', (c) => {
-  return c.text('Hello, world!');
+const testapi = new Hono();
+
+testapi.get('/', async (c) => {
```
andrew-bierman, line 40: Clean up or at least move to route file if you need it for debugging. I'm not against the idea of easy endpoints to quickly debug api status

server/src/controllers/item/importFromBucket.ts
andrew-bierman: Noticed you had service file already. I think that is good option, or a bucket integration layer with utils. Looking to keep controller as clean we can to minimize repeated code

andrew-bierman (changes requested):
Nice job, some file structure stuff and cleanup remaining

andrew-bierman:
@MuhammadHassan03 pls make new branch from here and rebase. @pinocchio-life-like's computer is in the shop

andrew-bierman (approved):
Nice work here, we need to update env example and the discord I believe

PR #1170, opened 2024-08-19
adiitional tests

JewelTee:


JewelTee:
@andrew-bierman yes, I think it's because the UI has been updated and some buttons were changed also. I'm working on it.

PR #1184, opened 2024-08-23
Configure remote appVersion management

mikib0:
Forgetting to update the [developer-facing build version](https://docs.expo.dev/build-reference/app-versions/#developer-facing-build-version)  could cause app store rejections. This will delegate incrementing of the build version to EAS Build; eliminating the need to manually increment it everytime when creating a new build.

https://docs.expo.dev/build-reference/app-versions

### DIFF
--- a/apps/expo/eas.json
+++ b/apps/expo/eas.json
@@ -1,6 +1,7 @@
 {
   "cli": {
-    "version": ">= 3.8.1"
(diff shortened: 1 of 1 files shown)


andrew-bierman (approved):
good call, this might solve the annoying xcode behavior of updating build # to always be higher than most recent in test flight

PR #1194, opened 2024-08-25
Fix app icon and name

mikib0:
**Before:**
before

**After:**
after


andrew-bierman (approved):

andrew-bierman:
Good call, let's use "PackRat" for app name

Issue #1200, opened 2024-08-27
Empty Cards Showing up Under Similar Packs


JewelTee:
**Describe the bug**
In the "Similar Packs" section some cards are blank and do not show any data. It would be better not to display empty cards. 

**To Reproduce**
1. Navigate to the page or section where similar packs are displayed.
2. Observe the first card, which displays data correctly.
3. Observe the other cards after the first one.

**Expected behavior**
All cards in the "Similar Packs" section should display relevant data with no additional blank cards.

**Screenshots**

**IOS**

https://github.com/user-attachments/assets/83cea727-3e51-46ec-89ea-a99859ad4432

**Android**

https://github.com/user-attachments/assets/d042432d-bd7a-4f66-9c9f-be8252515999

**Additional context**
The issue persists both on IOS and Android.

mikib0:
Did you observe this in the dev environment?

JewelTee:
> Did you observe this in the dev environment?

No, I observe this in the production build

andrew-bierman:
We should probably do no arrows on native and just allow for horizontal scroll. The arrows are good for web I think but make less sense in native, and we can gain some more screen real estate if we remove

mikib0:
I agree. Additionally the corousel doesn't scroll on swipe on native in my experience. Will work this.

mikib0:
Maybe we have a seperate issue to track that or in #1177, but this one is resolved.

PR #1205, opened 2024-08-29
Fix vectorize issues

mikib0:
Changes:
- Comment out vectorize binding configuration since REST api (instead of vectorize client) is used currently. This is to also fix the binding name conflicting with VECTOR_INDEX var which is used by the REST api --causes deployment failure.
- The current production vector index was used in both dev and prod. This leads to cluttering of the vector (with dev and prod data) and results in empty cards showing up under similar packs (see: #1200). This was fixed by using a different one for dev instead. The ultimate solution however is to change the production vector index to an entirely new one, thus eliminating the chance of cluttering even when running from previous versions (older commits).

### DIFF
--- a/server/wrangler.toml.example
+++ b/server/wrangler.toml.example
@@ -59,7 +59,7 @@ binding = "AI"
 
 #[[env.production.vectorize]]
 #binding = "VECTOR_INDEX"
-#index_name = "vector-index"
+#index_name = "vector-index-production"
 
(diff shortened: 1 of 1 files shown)


andrew-bierman (approved):

andrew-bierman:
We should clean prod vector db, and also maybe have a cleanup service running on cron to clear our stale vectors ?

PR #1209, opened 2024-08-31
Add biometric authentication 

Isthisanmol:


JewelTee (approved):
@Isthisanmol Biometric authentication works well on my end 👍🏻

packages/app/modules/auth/components/AuthWrapper.tsx
```diff
@@ -14,6 +16,60 @@ export const AuthWrapper = ({
...
+    if (result.success) {
+      setIsAuthenticated(true);
+    }
+  };
+
+  useEffect(() => {
+    authenticate();
+  }, []);
+
+  if (!isAuthenticated) {
```
pinocchio-life-like, line 55: I believe this should be either platform specific or it be set on web as well like using WebAuthn or so. 

or just bypass it on web cause this won't allow pages wrapped with the context to be accessed.

pinocchio-life-like (approved):

andrew-bierman (approved):

PR #1212, opened 2024-09-01
tamagui alerts 

mikeCyb:
The pack item delete modal has been updated to use the new alert component.


EngrAJ8888:
Hi @mikeCyb - I’m not seeing the alert dialog when I try to delete an entire pack. It only appears when deleting individual items. Thanks!

EngrAJ8888:
Hi @mikeCyb, we should also add an alert dialog box for deleting the whole pack. Other than that, I don’t think any other updates are needed. Thanks!

andrew-bierman (approved):

PR #1214, opened 2024-09-03
Make register and login screen consistent

Isthisanmol:


JewelTee (approved):
@Isthisanmol looks great!

PR #1217, opened 2024-09-04
Fix sending a bunch of getMe requests

mikib0:


andrew-bierman (approved):
We might need to double check our auth token logic to make sure logout and protected routes still work with this change

PR #1227, opened 2024-09-09
Feed infinte scroll

pinocchio-life-like:


packages/app/modules/feed/hooks/useFeed.ts
```diff
@@ -15,10 +24,10 @@ export const useFeed = ({
   feedType: string;
   selectedTypes: Object;
   id: string;
-}> = {}) => {
+}> = {}): UseFeedResult => {
   switch (feedType) {
     case 'public':
-      return usePublicFeed(queryString, selectedTypes);
+      return usePublicFeed(queryString, selectedTypes); // Use the typed return from usePublicFeed
```
andrew-bierman, line 30: Do we not get a warning about violating the rules of hooks here?

andrew-bierman (approved):
Check that switch statement with hooks and make sure that is good, might need to revise that flow.

Otherwise this looks good. Is that other PR still needed ?

PR #1230, opened 2024-09-10
Fix: login screen not scrollable

Tadjaur:
The login screen should automatically scroll and show the current input field when the virtual keyboard show up on mobile device.

This actual look:
image

The expected look:
image


andrew-bierman (approved):
Lgtm

PR #1237, opened 2024-09-14
Fix deployment issues

mikib0:


Tadjaur:
Wow, looks like we worked on the same problem. We should find a way to synchronize better in the future.

Tadjaur:
Very sorry But I think this pr should be close in favor of the following https://github.com/andrew-bierman/PackRat/pull/1239

Reasons:
- The pr [#1239](https://github.com/andrew-bierman/PackRat/pull/1239) has been tested on production and preview environments
- The pr [#1239](https://github.com/andrew-bierman/PackRat/pull/1239) change all content update by this pr and improve the deployment value
- This pr do not completely fix the backend deployment: We will still have an issue when uploading the secret because of some environment variable like `BUCKET_SESSION_TOKEN` that can have empty string.

mikib0:
@Tadjaur
> Wow, looks like we worked on the same problem.

Not quite, this PR also adds the new variables we have (part of the deployment issues I believe). But, yeah we should definitely close this for the reasons you outlined and then we update #1239 to include the new vars. Thanks for catching that :smile:

Tadjaur:
> @Tadjaur
> > Wow, looks like we worked on the same problem.
> 
> Not quite, this PR also adds the new variables we have (part of the deployment issues I believe). But, yeah we should definitely close this for the reasons you outlined and then we update #1239 to include the new vars. Thanks for catching that :smile:

What are those new variables? So I can update the other branch

mikib0:
@Tadjaur It's `REFRESH_TOKEN_SECRET`, but also we need to remove vectorize binding and update the variable name for vector index. I'll highlight those changes in review. Thanks.

Tadjaur:
Oh okay I understand why I missed this one, the variable is not yet defined in our GitHub secrets

Tadjaur:
Updated.

PR #1249, opened 2024-09-19
use expo router for chat on native

Isthisanmol:


JewelTee (commented):
@Isthisanmol I am unable to log in or register on native

JewelTee (commented):
@Isthisanmol  I was able to sign in but I need to create a new user first. The chat is also functioning well :)

taronaleksanian (approved):

PR #1261, opened 2024-09-22
Fix/map

taronaleksanian:


packages/app/components/destination/index.tsx
```diff
@@ -182,7 +174,7 @@ export const DestinationPage = () => {
               </View>
 
               <DestinationHeader
-                geoJSON={geoJSON}
+                geojsonProperties={geojsonData.properties}
```
taronaleksanian, line 177: in destination page we don't save anything to db so perhaps we don't need it

taronaleksanian:
@mikib0 we have unplanned changes with the latest commit, can you please revert that and include only changes that we have arranged, and make a new pr to this branch with planned changes

PR #1282, opened 2024-10-04
Native build UI enhancement

MuhammadHassan03:


packages/ui/src/Bento/forms/layouts/SignInScreen.tsx
```diff
@@ -47,7 +47,7 @@ export function SignInScreen({
         'Error',
         'No biometrics are enrolled. Please set up biometrics in your device settings.',
       );
-      return false;
+      return true;
```
taronaleksanian, line 50: is it fixes the ui of sign in screen?

PR #1285, opened 2024-10-04
Fix api workflow

mikib0:
Workflow breaks when there's a conflicting normal var vs secret. So need to watch out that secret vars aren't in env vars in wrangler config when deploying manually. It's easy to miss this though, thus I think again we need to revamp our dev workflow. For now, redeploying without including the secrets (incase they were mistakenly included) would fix the workflow  🤞.


taronaleksanian:
> Workflow breaks when there's a conflicting normal var vs secret. So need to watch out that secret vars aren't in env vars in wrangler config when deploying manually. It's easy to miss this though, thus I think again we need to revamp our dev workflow. For now, redeploying without including the secrets (incase they were mistakenly included) would fix the workflow 🤞.

so we'll have changes on our local setup right?, can you provide the working wrangler.toml or any other config file if we need so we can change it on our local setup?

mikib0:
> so we'll have changes on our local setup right?, can you provide the working wrangler.toml or any other config file if we need so we can change it on our local setup?

Okay sure thing.

taronaleksanian (approved):

PR #1355, opened 2024-12-11
Fix quantity on native

Isthisanmol:


taronaleksanian (approved):

JewelTee (approved):
functions well now, thank you!

PR #1390, opened 2024-12-26
Add search on native

Isthisanmol:


JewelTee (commented):
@Isthisanmol hi, The search icons are looking good and functioning well too. Just a suggestion, theres already a destination search field on the dashboard I think we should remove the search icon on that screen ? 
Screenshot_20241227_211010_Packrat

JewelTee (approved):

taronaleksanian (approved):

PR #1397, opened 2024-12-30
Development

taronaleksanian:


server/src/services/pack/getSimilarPacksService.ts
```diff
@@ -29,7 +29,9 @@ export async function getSimilarPacksService(
   const {
     result: { matches },
   } = await VectorClient.instance.search(pack.name, 'packs', limit, {
-    isPublic: true,
+    isPublic: {
+      $eq: true,
+    },
```
Tadjaur, line 34: @taronaleksanian does this change really fix the issue?

Tadjaur: I'm asking because according to the [documentation](https://developers.cloudflare.com/vectorize/reference/metadata-filtering/#implicit-eq-operator) it seems to be the same thing and when I test this locally , similar packs are still not displayed.

andrew-bierman (approved):

PR #1403, opened 2025-01-05
Type/fix app

pinocchio-life-like:


taronaleksanian (changes requested):
I'm getting a crash screen if I run the app on ios

PR #1453, opened 2025-06-22
Chore/Add Sentry

mikib0:


andrew-bierman (approved):
Looks good, see comment on special hono middleware. We should also go all in on either bugsnag or sentry - I'm fine either way

andrew-bierman (approved):
Lgtm

PR #1455, opened 2025-06-23
add similarity search

Isthisanmol:


andrew-bierman (approved):
lgtm - do we have corresponding expo PR ? also double check we are not duplicating the `generateEmbedding` utility. I dont think so but wanted to double check

mikib0 (approved):
Lgtm

PR #1471, opened 2025-06-30
Fix/Restore Proper Catalog Pagination

mikib0:


andrew-bierman (approved):
This looks good, thanks. 

We should move catalog item by id to dedicated route also, but lets get this in first

PR #1472, opened 2025-07-01
Feat/Enhance Packs & Items Retrieval

mikib0:


andrew-bierman:
> Yikes conflicts

I think if we just load up the service files with query building to keep the routes clean, conflicts won't be too bad ?

mikib0:
Totally. I've began restructuring in fact. Already cleaned up catalogs a bit -> #1474

PR #1488, opened 2025-07-03
Feat/GitHub packages

andrew-bierman:


mikib0:
You actually got this running on your end? I'm hitting module resolution error from the nativewindui package.
```bash
Android Bundling failed 3491ms node_modules/expo-router/entry.js (3076 modules)
Unable to resolve "nativewindui/lib/useColorScheme" from "node_modules/@packrat-ai/nativewindui/ActivityIndicator.tsx"
```

mikib0:
Probably metro can't compile ts ? I'll try precompiling see if that works.

mikib0:
I've set up transpilation and got it working with Metro, but hitting issues with events and some styles. Digging into that.

mikib0:
Nativewind definitely didn't play well with transpiled code and I couldn't figure out why. Turns out metro surprisingly works with source although it doesn't seem to respect tsconfig because path aliases break despite having them defined in the config. Updating those paths to relative imports got everything working.

mikib0 (approved):

PR #1490, opened 2025-07-09
Improve ETL: skip items with missing SKUs and purge bad data

Isthisanmol:
This PR fixes issues with misaligned catalog data by validating that each item has both a name and a non-empty SKU before processing. It also removes any existing items in the database with null or blank SKUs to clean up legacy bad data added by earlier spiders. This ensures that only valid, properly structured items are inserted or updated going forward.


andrew-bierman (approved):
Lgtm

PR #1492, opened 2025-07-10
Fix/Lint

mikib0:


mikib0:
There are still quite a few type errors - I’ve already fixed a bunch here, but more remain. Will keep cleaning things up as we go.

PR #1597, opened 2025-08-21
Fix ai suggestion bubble

### LINKED ISSUE #1588: BUG: AI Chat Suggested Options Display Incorrectly when clicked
**Description:**
In the AI Chat feature, pre-made options are provided at the start of the chat. When tapping one of these options, the app displays a small **blue text balloon** instead of showing the full text of the selected option.

**Steps to Reproduce:**

1. Open the AI Chat feature in the app.
2. At the start of the chat, tap one of the suggested options.
3. Observe the message bubble that appears.

**Screenrecording:**

https://github.com/user-attachments/assets/7ec9ac65-a0f2-4104-a1c7-3840f018f5c5

**Expected Result:**

* The chat should display the full text of the selected suggested option in a normal message bubble.

**Actual Result:**

* A small blue text balloon appears instead of the full option text.

Isthisanmol:
closes #1588

### DIFF
--- a/apps/expo/app/(app)/ai-chat.tsx
+++ b/apps/expo/app/(app)/ai-chat.tsx
@@ -133,8 +133,8 @@ export default function AIChat() {
   }, [error]);
 
   const handleSuggestionPress = (suggestion: string) => {
-    setInput(suggestion);
-    handleSubmit();
+    sendMessage({ text: suggestion });
+    setInput('');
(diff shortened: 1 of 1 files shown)


Isthisanmol:
> lgtm, if messages is valid in dep array

Yes, messages is required in the dependency array for better ux.

PR #1653, opened 2025-08-25
Fix/Patch Image Caching When Adding Catalog Item to Pack 

### LINKED ISSUE #1589: BUG: Catalog Item Image Not Displayed After Adding to Pack
**Description:**
When adding an item from the catalog that includes an image, the item successfully adds to a pack, but the image does not display in the pack view.

**Steps to Reproduce:**

1. Open the app and navigate to the Catalog.
2. Select an item that has an image.
3. Add the item to a pack.
4. Open the pack and observe the item.

**Screenrecording:**

https://github.com/user-attachments/assets/4121d4ba-5bbb-4f0e-806a-5824584bcee9

**Expected Result:**

* The item image should display correctly in the pack.

**Actual Result:**

* The item is added, but no image is shown in the pack.

mikib0:
fixes #1589


mikib0:
Apparently copilot broke the caching logic.

Issue #1652, opened 2025-08-25
AI Chat – Slow Response, Missing Auto-Scroll, and No Chat Bubble


JewelTee:
**Description:**
When using the AI chat feature, several issues occur:

1. The AI response loads very slowly.
2. The chat does not auto-scroll as the AI types, requiring the user to manually scroll down to see the latest response.
3. The UI for AI messages is inconsistent — user messages appear in chat bubbles, but AI responses do not have a bubble, making the interface look unfinished.

**Steps to Reproduce:**

1. Open the AI chat feature.
2. Ask a question to trigger an AI response.
3. Observe the response speed, scrolling behavior, and UI display.

**Screenrecording:**

https://github.com/user-attachments/assets/4db48f80-ff88-4262-bc4c-be2907fc4377

**Expected Behavior:**

* The AI should respond smoothly and in a timely manner.
* The chat should auto-scroll to show the newest messages without manual intervention.
* Both user and AI messages should appear in consistent chat bubbles for a clean UI.

**Actual Behavior:**

* AI responses load very slowly.
* Chat does not auto-scroll automatically.
* AI messages are missing chat bubbles.

App Version: Latest TestFlight build 2.0.5

mikib0:
@JewelTee That was on me!🙏

JewelTee:
> > The chat should auto-scroll to show the newest messages without manual intervention.
> 
> Do you mean it should scroll as it's streaming?

For me, I think yes. But if this if it's intended to not automatically scroll then its not an issue 😅

mikib0:
AI response is definitely slow. Took almost 20secs here! Do you have a feeling that it used to be faster @JewelTee ? 

I'm trying to rule out CF AI gateway.

Isthisanmol:
@JewelTee Trying to add auto-scroll.

JewelTee:
> AI response is definitely slow. Took almost 20secs here! Do you have a feeling that it used to be faster [@JewelTee](https://github.com/JewelTee) ?
> 
> I'm trying to rule out CF AI gateway.

Yeah,  AI responses used to be faster and more seamless

JewelTee:
[...]

PR #1663, opened 2025-08-28
Add scroll button

Isthisanmol:
Discussed in #1652.
Currently, the scroll button does not appear or work as soon as the streaming text goes below the screen, because I think chat streaming freezes the screen. If we show the button immediately after the text goes off-screen, it won’t be clickable and will look buggy. So for now, I’ve added it to appear only when the status is ready.


mikib0 (approved):
lgtm!

PR #1667, opened 2025-08-29
Fix/cleanup ai

mikib0:
#1671 

Also addressed in this PR:
- suggestions going under greeting message
- report btn appearing before generation is complete
- awkward layout issues especially during stream
- standardized tool call output


mikib0:
@copilot please continue this PR by finishing the remaning tool calls. Follow the same pattern as the completed tool calls in this PR.

PR #1733, opened 2025-09-30
Feat/expo 54

andrew-bierman:


mikib0:
Slowing down on this because the crashing issue persits despite multiple attempts. I'll try out other fixes on the side while I shift focus to something else. I'm not sure if you guys also face those issues so far? The root cause isn't exactly clear yet as it just crashes silently but i'm suspecting breaking changes in `react-native-reanimated` v4 affecting our nwui dep. I tried downgrading to v3 but it doesn't seem to compatible with React Native 0.81 (?) because build is failing in that case with complation errors.

PR #1737, opened 2025-10-11
Implement trip feature

Isthisanmol:


apps/expo/features/trips/components/TripForm.tsx
mikib0: @Isthisanmol Date picker opens automatically in trip form. Also it takes multiple presses on cancel or ok to dismiss. Have you noticed this on your side as well?

mikib0 (approved):
Please check out that pending comment when you are able. Also it looks like biome check is not passing. Otherwise everything lgtm!

PR #1766, opened 2025-11-11
Chore/Resolve Test Failures

mikib0:


mikib0:
@copilot resolve pack template test failures. tests expect 200 but are receiving 404. we might likely need fixtures and helper functions for creating test pack templates as in the case with catalog.

PR #2045, opened 2026-04-11
chore(lint): harden linting with additional rules and custom checks

andrew-bierman:
## Summary

Tightens linting across the monorepo to enforce code-quality rules the team has discussed.

### Biome rules enabled

| Rule | Severity | Existing violations | Status |
|---|---|---|---|
| `correctness/noUnusedImports` | `error` (was recommended warn) | 3 | fixed in this PR |
| `style/noNonNullAssertion` | `error` (was recommended warn) | 8 | fixed in this PR |
| `complexity/noForEach` | `warn` | 45 | pending cleanup |
| `performance/useTopLevelRegex` | `warn` | 98 | pending cleanup |

The two rules promoted to `error` had all their violations fixed in this PR — non-null assertions were replaced with `assertDefined()` from `@packrat/guards` (or the analytics-local `type-assertions` module), or converted to optional chaining where the call site was already safe. Unused imports were auto-removed by Biome's `--unsafe` fixer.

Biome `recommended` already covered `noExplicitAny` (152), `useOptionalChain`, `useNumericLiterals`, `noDoubleEquals`, `useValidTypeof`, `useExhaustiveDependencies`, and friends, so they're already active at their recommended severities.

### Custom checks added

Two opt-in shell scripts enforce project conventions Biome can't express:

- `scripts/lint/no-raw-typeof.sh` — flags `typeof x === 'string' | 'number' | ...` in anything outside of the `@packrat/guards` package (which implements the guards) and test files. The team has agreed that primitive narrowing should go through `isString` / `isNumber` / etc. from `@packrat/guards`. **Currently reports 66 violations.**
- `scripts/lint/no-raw-regex.sh` — flags `new RegExp(...)` and regex literals passed directly to string methods (`.replace`, `.match`, `.test`, `.split`, `.search`, `.replaceAll`) in non-test production code. The `magic-regexp` reference lives in `packages/analytics/src/core/enrichment.ts`. **Currently reports 91 violations.**

Both use POSIX `grep` + `find` so they work in CI without extra tooling (no ripgrep dependency) and they skip `node_modules`, `dist`, `build`, tests, and the `enrichment.ts` reference implementation.

New npm scripts:

- `bun lint:custom` — runs both shell scripts sequentially.
- `bun lint:strict` — runs `biome check` followed by `lint:custom`.

Neither is wired into the default `bun lint` or CI yet. The backlog is too large to fix in one PR, so `lint:strict` ships as an opt-in developer tool and a future-CI target.

### Rules NOT enabled (and why)

- **`style/noMagicNumbers`** — 1432 existing violations across the monorepo. Far too noisy to ship as `warn`, let alone `error`. Would need to be introduced with a `files.includes` scoped to a handful of paths first, or pared down with a bigger allowlist. Deferred.
- **`suspicious/noConsole`** — 480 existing violations, mostly in scripts, tests, and ad-hoc logging in services. Would need per-package overrides (`scripts/**` and tests should stay unrestricted). Deferred.
- **"No magic strings"** — Biome doesn't have a rule for this; enforcing "string literals used as enum-like values must come from a const/enum" generically would require an AST walker with domain-specific hints. Deferred.

### Rules enabled to `warn` pending cleanup

- `complexity/noForEach` — 45 violations. Easy mechanical replacements (`.forEach` → `for…of`), good candidate for a follow-up PR.
- `performance/useTopLevelRegex` — 98 violations. Needs judgment per call site; some will become `const PATTERN = createRegExp(...)` using `magic-regexp` (aligning with the new `no-raw-regex.sh` push).

## Migration path

Follow-up PRs can:

1. Fix the `warn`-level Biome violations file-by-file (noForEach, useTopLevelRegex).
2. Migrate the 66 `typeof` call sites to `@packrat/guards` so `no-raw-typeof.sh` can go green.
3. Migrate the 91 regex call sites to `magic-regexp` so `no-raw-regex.sh` can go green.
4. Promote `warn` → `error` for cleaned-up rules.
5. Wire `lint:strict` into `biome.yml` CI workflow.

## Test plan

- [x] `bun biome check` exits 0 (315 warnings, 0 errors)
- [x] `bun run check-types` passes
- [x] `bash scripts/lint/no-raw-typeof.sh` exits 1 with 66 violations (expected — the rule is opt-in)
- [x] `bash scripts/lint/no-raw-regex.sh` exits 1 with 91 violations (expected — the rule is opt-in)
- [x] Pre-commit hook (`bun check {staged_files}`) runs on every commit in this PR without blocking
- [ ] CI green on the PR


andrew-bierman:
@copilot custom scripts should be in typescript

PR #2212, opened 2026-04-16
Fix/Header-title-missing

Isthisanmol:


andrew-bierman:
- [ ] I'm hesitant to lose out on the native iOS search bar. Wondering if there is something else we can do / raise in react native screens

Isthisanmol:
Yeah, that’s fair. We can raise this issue in react-native-screens. Do you think I should create a minimal reproduction repo for it, or would sharing the Packrat repo be okay?

Isthisanmol:
I’ve added a workaround for the catalog, and everything is working as expected on my end. Now moving on to Guides screen.

Also, we don’t need to lose the native iOS search bar.

mikib0 (approved):
Lgtm!

PR #2342, opened 2026-04-27
fix(admin): fall back to Bearer JWT when CF Access is disabled but env vars remain

andrew-bierman:
## Summary
- `adminAuthGuard` now falls back to Bearer JWT check when `CF_ACCESS_TEAM_DOMAIN`/`CF_ACCESS_AUD` are set but the request has no CF Access JWT
- Previously, any request without a valid CF Access JWT was immediately rejected with 401 when CF env vars were configured — breaking Basic-auth login after disabling CF Access on the dashboard

## Root Cause
When a user disables CF Access from the Cloudflare dashboard (but hasn't yet cleared the API env vars), requests arrive without a CF JWT. The guard treated the presence of env vars as a hard requirement for CF auth, ignoring the admin Bearer JWT issued by `/token`.

## Behaviour After Fix
| Scenario | Before | After |
|---|---|---|
| CF Access active, CF JWT present | ✅ allowed | ✅ allowed |
| CF Access disabled, env vars still set, Bearer JWT | ❌ 401 | ✅ allowed |
| No CF vars, Bearer JWT | ✅ allowed | ✅ allowed |
| No CF vars, Basic auth | ✅ allowed | ✅ allowed |

## Test plan
- [ ] Log in via Basic auth while `CF_ACCESS_TEAM_DOMAIN`/`CF_ACCESS_AUD` are set but no CF Access cookie is present — should succeed
- [ ] Log in via Basic auth with no CF env vars — still works
- [ ] With CF Access active, CF JWT still accepted

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added trail search and discovery functionality with spatial filtering and sport-based categorization

* **Improvements**
  * Enhanced environment variable validation with centralized schemas
  * Strengthened authentication and access control logic
  * Improved type safety and error handling throughout the application

* **Chores**
  * Extended CI/CD pipeline with additional custom linting checks and type-cast validation


andrew-bierman:
Closing — the original strict CF-only guard is correct. Root cause is getCFAccessIdentity() returning null after valid CF session; fixing that instead.

PR #2366, opened 2026-05-01
feat(web): web support MVP — platform shims, Playwright e2e, OTP fix

andrew-bierman:
## Summary

- **Platform shims** (`lib/*.ts` + `lib/*.web.ts`): `Picker`, `appleAuthentication`, `updates`, `devClient` — replaced 4 Metro-level stubs with proper platform abstractions
- **Web layouts**: `app/_layout.web.tsx` and `app/(app)/(tabs)/_layout.web.tsx` (no NativeTabs on web)
- **Store/atom web shims**: `atomWithSecureStorage.web.ts`, `uploadImage.web.ts`, `localModelManager.web.ts`, `providers/index.web.tsx`
- **Playwright e2e suite** (`apps/expo/playwright/`): auth, packs, trips, profile flows with CI workflow
- **Bug fixes**:
  - OTP: `orderBy desc(expiresAt)` to fetch newest code, not oldest
  - `trips/index.ts`: suppress deleted-trip 404s  
  - `one-time-password.tsx`: guard `setNativeProps` on web
  - `profile/index.tsx`: `expo-updates` → `lib/updates`; `DataItem` gets optional `testID` field

## Test plan

- [ ] `bun check-types` passes
- [ ] iOS/Android build unaffected (lib/ native files re-export originals)
- [ ] Web boots without crashes (Metro no longer stubs expo-updates, dev-client, apple-auth, picker)
- [ ] OTP login delivers newest code correctly
- [ ] Playwright suite runs: `bun test:web` from `apps/expo`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added full web E2E test infrastructure and runnable test scripts.
  * Enabled trip edit and delete flows with confirmation dialogs.

* **Bug Fixes**
  * Improved web image sizing and layout for many screens.
  * Fixed OTP focus handling and stabilized auth navigation timing.

* **Tests**
  * Added Playwright E2E suites for core flows, packs, trips, profile, and more; test artifacts ignored by git.

* **Chores**
  * Added web stubs/polyfills and expanded test instrumentation (testIDs) across UI.

---

## Update — web auth now works cross-origin + reliable local e2e

This branch was extended to make the web app actually authenticate against the Workers/Elysia/**Better Auth** API and to make the web-e2e suite runnable.

**Web auth (was silently broken on web):**
[...]

PR #2374, opened 2026-05-01
poc(ui): shadcn/Radix web shims for nativewindui via Metro .web.tsx

andrew-bierman:
## Summary

- Adds `.web.tsx` shim files for every component in `packages/ui/nativewindui/` so the Expo web build picks up shadcn/Radix equivalents via Metro's platform extension resolution (`.web.tsx` > `.tsx`) — zero changes to screen code
- Extracts a shared `cn.web.ts` utility (clsx + tailwind-merge) used by all shims
- Extracts `use-color-scheme.web.ts` to break a circular import with `theme-toggle`
- Bridges the `Toggle` prop mismatch (`value`/`onValueChange` → `checked`/`onCheckedChange`)
- All findings from a full `ce:review` multi-agent pass addressed (16 todos, all resolved)

## Components shimmed

| nativewindui component | Web implementation |
|---|---|
| `Button` | `@packrat/web-ui` shadcn `Button` with variant/size mapping |
| `Toggle` | Wrapper over shadcn `Switch` bridging `value`/`onValueChange` |
| `Picker` / `PickerItem` | shadcn `Select` + `SelectItem` |
| `SearchInput` | Bare HTML `<input type="search">` + Lucide icon |
| `TextField` | HTML `<input>` / `<textarea>` with secure + multiline guard |
| `ActivityIndicator` | CSS spin animation via `<output aria-live="polite">` |
| `Icon` | Lucide-react icon map with Circle fallback |
| `Text` | Semantic HTML passthrough (`h1`–`h6`, `p`, `span`) |
| `List` / `ListItem` | `<ul>` / `<li>` passthrough |
| `LargeTitleHeader` | Sticky `<header>` with left/right view slots |
| `SegmentedControl` | Accessible `<div role="tablist">` with tab buttons |
| `Stepper` | `+`/`−` buttons with value display |
| `ThemeToggle` | System-theme aware sun/moon toggle |
| `useColorScheme` | `next-themes` `useTheme` adapter |

## Key decisions

- **No screen code changes** — Metro's `.web.tsx` extension resolution does all the heavy lifting
- **`@packrat/web-ui`** promoted from `devDependencies` → `dependencies` in `packages/ui/package.json` (runtime import, not just type reference)
- **`@packrat/guards` `isFunction`** used in place of raw `typeof` checks (pre-push hook enforcement)
- **tsconfig exclude** extended to cover `*.web.ts` alongside `*.web.tsx` to keep new shared files out of native compilation

## Testing

- Biome lint: ✅ clean
- Pre-push hooks (`clean-checks`, circular-dep scan, type-cast scan): ✅ all pass
- Manual: shims render correctly on web dev server

## Post-Deploy Monitoring & Validation

No additional operational monitoring required: these are build-time shim files with no runtime data path or server-side effects. Any regression would be visible immediately on the web dev server.

---

[Compound Engineered](https://github.com/EveryInc/compound-engineering-plugin) 🤖 Generated with [Claude Code](https://claude.com/claude-code)


andrew-bierman:
Closing this POC in favor of #2366 as the chosen web-support direction.

After comparing the two approaches:
- **#2374 (this PR)**: Metro `.web.tsx` shims swapping nativewindui → shadcn/Radix component-by-component. 87 files. Lower-risk on a component basis, but creates a parallel UI surface that has to be kept in sync with every nativewindui change going forward.
- **#2366**: Platform shims for native APIs (Picker, AppleAuthentication, Updates, DevClient) + targeted web layouts + atom/store shims. 79 files. Keeps nativewindui as the single source of truth for components and only shims what's genuinely platform-divergent.

The maintenance cost of #2374's parallel UI library is what tips it — every nativewindui upgrade would require us to also touch each shadcn shim. #2366 only diverges where the platform truly differs (native modules), so nativewindui upgrades flow through unmodified.

The diff here is preserved on `poc/shadcn-web-shims` if we want to mine specific shims (e.g. components where nativewindui's web behavior is genuinely broken and a shadcn fallback is better than fixing in-place).

If/when nativewindui needs per-component web overrides, we can do that incrementally on top of #2366's foundation rather than as a parallel UI track.

PR #2422, opened 2026-05-14
chore(lint): enforce single-parameter owned functions

andrew-bierman:
## Summary

- Restores `scripts/lint/no-owned-max-params.ts` — a custom check lost in a force-push reset
- Biome stays at `useMaxParams: max 2` as a general backstop; this script adds the project-specific rule: **owned function definitions take at most 1 parameter**; inline framework callbacks (renderItem, onChange, etc.) are excluded
- Fixes all 206 violations by converting 2-param owned functions to the `(subject, opts)` / destructured-object shape — no suppressions, no biome-ignore, no ts-ignore

## Commits

1. `chore(lint): restore no-owned-max-params custom check script`
2. `refactor: convert all owned functions to single object-param signatures` (206 violations → 0)
3. `fix(casts): move safe-cast annotations inside object literals to satisfy lookback check`

## Post-Deploy Monitoring & Validation

No additional operational monitoring required: pure lint/style enforcement with no runtime impact.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2429, opened 2026-05-16
chore(lint): enforce single-parameter function style with custom check

andrew-bierman:
### Motivation
- The repository lint baseline (`biome`) is set to allow up to 2 parameters, but the team wants to tighten style to prefer a single typed object parameter for first-party code. 
- Introduce a stricter, repo-specific check to enforce a one-parameter rule while avoiding noise from callbacks and third-party code. 
- Provide an explicit opt-out when multiple parameters are genuinely required to make incremental migration feasible.

### Description
- Added a new custom check script `packages/checks/src/check-max-params.ts` which scans `apps/` and `packages/` TypeScript files and reports any function-like node with more than one parameter.  
- The checker ignores common test/build/generated paths, excludes `node_modules`/dist/build, and treats callbacks passed directly into call expressions as non-violations to reduce false positives.  
- Introduced an escape hatch annotation `@allow-multi-param` that can be placed on a leading comment to exempt specific functions.  
- Wired the new script into npm scripts: added `check:max-params` to `packages/checks/package.json` and a root-level `check:max-params` and appended it to the `lint:custom` pipeline in the root `package.json` while leaving Biome's `useMaxParams` at `max: 2` as the baseline.

### Testing
- Ran `bunx biome format --write packages/checks/src/check-max-params.ts` which formatted the new file successfully.  
- Ran `biome check --no-errors-on-unmatched` against the changed files as part of the pre-commit hook which succeeded after formatting.  
- Ran `bun run --cwd packages/checks check:max-params` which intentionally failed because it reported existing multi-parameter functions in first-party code (this is expected and indicates the check is detecting current violations).

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a07baa6115c8323b40992d96e85142c)

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "ios": "cd apps/expo && bun ios",
     "lefthook": "lefthook install",
     "lint": "biome check --write",
-    "lint:custom": "bun run scripts/lint/no-raw-typeof.ts && bun run scripts/lint/no-raw-regex.ts && bun run packages/env/scripts/no-raw-process-env.ts && bun run scripts/lint/no-duplicate-guards.ts && bun run scripts/lint/no-unauth-routes.ts",
+    "lint:custom": "bun run scripts/lint/no-raw-typeof.ts && bun run scripts/lint/no-raw-regex.ts && bun run packages/env/scripts/no-raw-process-env.ts && bun run scripts/lint/no-duplicate-guards.ts && bun run scripts/lint/no-unauth-routes.ts && bun run check:max-params",
     "lint:strict": "biome check && bun run lint:custom",
     "lint-unsafe": "biome check --write --unsafe",
     "mcp": "bun run --cwd packages/mcp dev",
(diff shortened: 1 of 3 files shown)


andrew-bierman:
Closing as superseded by #2422.

After comparing the two checks, #2422's `scripts/lint/no-owned-max-params.ts` is the better-designed implementation:

- Distinguishes **owned** function definitions from inline framework callbacks (e.g. `fetch`, `queue`, `resolveRequest` on Cloudflare Workers; React event handlers like `onChange`/`keyExtractor`); this check's heuristic is broader and would flag legitimate framework signatures.
- Has explicit per-file exclusions for cases that intentionally mirror external APIs (e.g. `packages/api/src/services/r2-bucket.ts` mirrors R2's positional API).
- Documents the why inline so the next contributor understands the rationale.

#2422 also ships the **fix** alongside the check (206 violations refactored across 260 files), which is what unblocks the check from being red on day one. #2429 only added the check without addressing existing violations.

If we later want the `@allow-multi-param` escape-hatch annotation from this branch, it's a small follow-up commit on top of #2422's check rather than a separate parallel implementation. Filing as docs/plans follow-up if useful.

Branch `codex/update-lint-rules-for-max-params` is preserved for reference.

PR #2431, opened 2026-05-16
api: implement chat rate-limiter and enable rate-limiting tests

andrew-bierman:
### Motivation
- The `/chat` route had a TODO around rate limiting and the corresponding tests were skipped, leaving an unimplemented production concern and no coverage for concurrent request throttling.

### Description
- Extract `TOKEN_RATE_LIMITER` from `getEnv()` and accept the `request` object in the `POST /chat` handler. 
- Apply runtime throttling when the optional `TOKEN_RATE_LIMITER` binding is present, composing the limiter key as `chat:${user.userId}:${ip}` and returning `429` when quota is exhausted. 
- Leave AI provider and streaming behavior unchanged when the request is allowed. 
- Re-enabled the previously skipped rate-limiting test block in `packages/api/test/chat.test.ts` so the endpoint behavior is covered by tests.

### Testing
- Ran the package unit test suite with `bun run test:api:unit` and all `packages/api` unit tests passed, including the newly active rate-limiting test. 
- The `chat` tests assert that concurrent requests return either `200` or `429` as expected.

------
[Codex Task](https://chatgpt.com/codex/cloud/tasks/task_e_6a07c00a28788323979038c1a21d5c2f)

### DIFF
--- a/packages/api/src/routes/chat.ts
+++ b/packages/api/src/routes/chat.ts
@@ -22,7 +22,7 @@ export const chatRoutes = new Elysia({ prefix: '/chat' })
   // Chat streaming
   .post(
     '/',
-    async ({ body, user }) => {
+    async ({ body, user, request }) => {
       const typedBody = body as {
         messages?: UIMessage[] | undefined;
         contextType?: string;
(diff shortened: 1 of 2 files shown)


andrew-bierman:
Verified: `TOKEN_RATE_LIMITER` binding exists in `packages/api/wrangler.jsonc` for both prod (top-level, line ~40) and dev env (line ~133), so this won't silently no-op in either environment. Wire-up looks clean — `request.headers.get('cf-connecting-ip')` keyed alongside `user.userId` is the right shape. One ask before merging: please confirm the test fixture mocks the rate-limit binding so CI doesn't depend on a real RL service. Otherwise marking ready.

PR #2457, opened 2026-05-20
fix(ci): fix failing unit tests and Playwright global setup

mikib0:
## Summary

- **`passwordResetService.test.ts`**: Mock for `timingSafeEqual` used positional args `(a, b)` but the real function takes `{ a, b }` — the mock always returned `false`, making all OTP-comparison-dependent tests unreachable
- **`embeddingHelper.test.ts`**: 7 fallback tests called `getEmbeddingText(item, existingItem)` as positional args instead of the `{ item, existingItem }` object form, causing `TypeError: Cannot read properties of undefined (reading 'name')`
- **`getRelativeTime.test.ts`**: 3 i18n tests called `getRelativeTime(date, t)` positionally instead of `{ dateValue, t }`, so the translate spy was never invoked
- **`globalSetup.ts`** (Playwright): Login endpoint `/api/auth/login` was removed in the Better Auth migration — replaced with `/api/auth/sign-in/email` and updated response destructuring to read `session.token` as the access token

## Test plan

- [x] `bun test:api:unit` — 299 tests pass (was: 12 failures in passwordResetService + embeddingHelper)
- [x] `bun test` (Expo) — 357 tests pass (was: 1 failure in getRelativeTime)
- [ ] Web E2E (Playwright) — will validate in CI; global setup now hits the correct login endpoint

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a web-specific API client for improved web auth behavior.
  * Web providers now include bottom-sheet modal support.

* **Bug Fixes / UX**
  * Trip creation no longer blocks when location is missing.
  * Confirmation dialog’s destructive button now labeled “Delete” and styled destructively.

* **Tests**
  * Many tests updated to object-style args, with stronger assertions and expanded E2E auth/context handling.

* **Chores**
  * Increased Playwright/CI timeouts and adjusted lint/test tooling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/PackRat-AI/PackRat/pull/2457?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/.github/workflows/web-e2e-tests.yml
+++ b/.github/workflows/web-e2e-tests.yml
@@ -57,7 +57,7 @@ jobs:
     needs: e2e-gate
     if: needs.e2e-gate.outputs.ready == 'true'
     runs-on: ubuntu-latest
-    timeout-minutes: 30
+    timeout-minutes: 60
 
     env:
       # The E2E user is upserted into the dev DB by the seed step below,
--- a/apps/expo/features/trips/store/trips.ts
+++ b/apps/expo/features/trips/store/trips.ts
@@ -14,9 +14,6 @@ const listTrips = async () => {
 };
 
 const createTrip = async (tripData: TripInStore) => {
-  if (!tripData.location) {
-    throw new Error('Trip location is required before sync');
-  }
   const { data, error } = await apiClient.trips.post({
     id: tripData.id,
     name: tripData.name,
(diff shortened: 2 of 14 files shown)


andrew-bierman:
Reviewed and attempted a merge against current development. After resolving the conflict set toward the current code, the remaining substantive diff is mostly superseded by newer work:

- Current development already has the Better Auth web E2E storage-state flow in globalSetup/fixtures.
- Current development already has the BottomSheet/action-sheet web provider structure and the object-arg test migrations that this PR originally fixed.
- The remaining new apps/expo/lib/api/packrat.web.ts would now override the newer unified apps/expo/lib/api/packrat.ts web path, dropping the current token cache/getApiBaseUrl behavior. That would be a regression risk.

Given that, I’m closing this as superseded rather than pushing a stale merge commit. The only remaining low-risk bit is longer Web E2E timeouts, which should be reintroduced separately if we still want it.

PR #2509, opened 2026-05-27
fix(cached-image): resolve race condition, error state with tap-to-retry, and corrupted cache prevention

mikib0:
## Summary

- **Race condition (B2):** Added a `cancelled` flag with effect cleanup so `setState` is never called after `imageObjectKey`/`imageRemoteUrl` change mid-flight or after unmount
- **Silent error state (B3):** Replaced the unresolved `// TODO` with a proper `hasError` state; on failure a tappable refresh-icon placeholder is rendered so users can retry the download in-place without navigating away
- **Corrupted cache (B4):** `ImageCacheManager.cacheRemoteImage` now deletes the partial file written by `downloadAsync` on non-200 responses — previously the corrupted file would be returned as a cache hit on the next mount, permanently breaking the image
- **Sentry (R1):** Added breadcrumbs before cache check and download, and `captureException` in the catch block per project conventions
- Also cleaned up: `undefined` className concatenation, redundant `bg-muted`/inline-style conflict on the loading placeholder, variable shadowing, and missing `imageRemoteUrl` guard

## Test plan

- [ ] Open a pack item detail screen — image loads normally (cache hit and miss paths both work)
- [ ] Disable network, open a pack item with no cached image — loading spinner appears then transitions to refresh icon placeholder
- [ ] Tap the refresh icon — spinner reappears, then image loads once network is restored
- [ ] Scroll a list quickly so `imageObjectKey` changes while a download is in-flight — no wrong image shown (validates race condition fix)
- [ ] Check Sentry breadcrumbs appear for image load events in dev


mikib0:
Closing in favour of a clean rebase on development — see replacement PR.

PR #2531, opened 2026-06-01
♻️ refactor: monorepo dependency policy + override audit

andrew-bierman:
## Summary

The repo carried two "shared-version" primitives — `catalog:` and root `overrides` — with no written rule for which to use, and an `overrides` block nobody could fully justify. This establishes a **dependency policy**, applies it to clear out overrides that no longer earn their place, and adds a lint so the policy can't silently rot.

**Result: 5 root overrides → 1.** Only `react` (the single-copy invariant) remains, documented with a removal condition that CI enforces.

## What changed

| | |
|---|---|
| **Policy** | `docs/dependency-policy.md` — when to use workspace / `catalog:` / `overrides`, plus a machine-checked fenced-JSON registry that justifies each surviving override with a removal condition. |
| **Override cleanup** | Dropped `@sinclair/typebox`, `elysia`, `expo-sqlite` (verified redundant against their origin commits), and `@packrat-ai/nativewindui` (resolved at the source — see below). |
| **Enforcement** | `check:overrides` lint (`scripts/lint/no-undocumented-overrides.ts`) wired into lefthook, `check-all`, and `checks.yml`. Adding an undocumented override now fails CI. |

### Why each dropped override was safe

- **`typebox`** — added to give elysia 1.4.x its required typebox `>=0.34.0` (`Unsafe` export). `packages/api` now pins `^0.34.15` directly, so elysia resolves `0.34.49` with `Unsafe` without the override. (RN's bundled jest tooling keeps its own isolated `0.27.10` in a separate subtree — expected, not a runtime duplicate.)
- **`elysia`** — added to prevent duplicate installs; superseded by the later catalog migration. Single elysia node confirmed after removal.
- **`expo-sqlite`** — an Expo SDK 55 pin; the direct `~55.0.15` in `apps/expo` holds it. Single node after removal.
- **`@packrat-ai/nativewindui`** — pinned a fork prerelease (`2.0.3-2`). The `2.0.6` release had an over-tight `react-native-keyboard-controller` peer floor (`^1.21.0`) that conflicted with Expo SDK 55's `1.20.7`. Rather than work around it, we **fixed it at the source** (PackRat-AI/nativewindui#19 → `2.1.0`, permissive `>=` peer ranges). `packages/ui` now pins `2.1.0` directly; keyboard-controller stays at the SDK's `1.20.7`, so there's no SDK divergence and no `expo.install.exclude` hack.

Each removal was checked via **version-node uniqueness** (one resolved copy in the tree), not just `bun pm why` requirer listings.

## Verification

- `bun check-types` ✓ · `bun test:api:unit` (328) ✓ · `bun test:expo` (357) ✓ · `bun test:lint` (16) ✓
- `bun check:overrides` ✓ (1 documented override) — including a live negative test (an undocumented/stale override exits 1)
- `expo-doctor` does **not** flag keyboard-controller (it's at the SDK's `1.20.7`)
- Multi-persona code review + GitHub Copilot review addressed

## Out of scope (deferred)

- **`container_src` build strategy** (workspace-ify vs `bun build --compile`) — coupled to the Turborepo migration and the `@tobyg74/tiktok-api-dl` decision; tracked in the brainstorm doc.

## Post-Deploy Monitoring & Validation

Dependency-graph + lint-config change. The one behavioral surface is the `@packrat-ai/nativewindui` `2.0.3-2` → `2.1.0` bump (a peer-range fix, same component code path). After merge:
- **Watch:** the first `apps/expo` build/preview after merge, for any NativeWindUI regressions.
- **Healthy signal:** Expo app renders NativeWindUI components as before; `checks` + `unit-tests` green; `bun check:overrides` passes.
- **Rollback trigger:** a regression traced to nativewindui `2.1.0` → revert the adoption commit (restores `2.0.3-2` + its override).
- **Note:** `development` already has an unrelated `expo-doctor` patch-version-drift finding (e.g. `expo-router`); not introduced here.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)
Claude Code

## Summary by CodeRabbit

* **New Features**
  * Added a policy for managing dependency versions and overrides, including a documented registry for override notes and removal conditions.
  * Introduced a new automated check that verifies overrides are documented and stay in sync with the policy.

* **Bug Fixes**
  * Improved validation in local and CI checks to catch missing, stale, or incomplete override entries before changes are merged.

* **Documentation**
  * Added planning and policy docs describing how dependency decisions should be recorded and reviewed.


andrew-bierman:
Thanks @copilot — both findings were valid and are addressed in `cb1c7a4e9`:

1. **nativewindui 2.0.6 peer conflict** — confirmed: `2.0.6` requires `react-native-keyboard-controller ^1.21.0`, but Expo SDK 55 pins `1.20.7` (and `2.0.3-2` only required `^1.16.7`, which the SDK satisfies). The expo test suite passed but doesn't validate peer ranges, so it slipped through. **Reverted the bump** back to `2.0.3-2` and kept the override, documented with this exact conflict as its `removeWhen`.
2. **typebox "exactly one" criterion** — scoped the plan's uniqueness criterion to the API/Elysia runtime copy and named the expected isolated `@jest/schemas/@sinclair/typebox@0.27.10` (RN's bundled jest tooling) as a correct, separate-subtree copy.

andrew-bierman:
Thanks @coderabbitai — both addressed in `d4fad40ed`:

1. **`no-undocumented-overrides.ts`** — changed the capture return to `match?.[1] ?? null` for `noUncheckedIndexedAccess` safety.
2. **Plan doc end-state** — added a post-execution outcome note recording the divergence from the planned 5→2 to the final 5→1 (the `@packrat-ai/nativewindui` override was also dropped after adopting `2.1.0`, whose permissive peer ranges fixed the keyboard-controller conflict). The authoritative end-state remains `docs/dependency-policy.md` (registry: `react` only).

PR #2554, opened 2026-06-02
🐛 fix(api): vector search uses HNSW index (was bypassing it, ~115× cost)

andrew-bierman:
## Summary

The 14 GB HNSW index on \`catalog_items.embedding\` has **0 lifetime scans** — verified against prod via \`pg_stat_user_indexes\` (\`stats_reset\` is null, so this measures the database's entire lifetime). Every vector search since the feature shipped has been doing **Seq Scan + parallel-Sort over 1.79M rows × 1536-dim cosine distance** instead of an HNSW top-k traversal.

**Root cause:** Drizzle's \`cosineDistance(...)\` wrapped in \`1 - (...)\` for the \`similarity\` value, then ordered by \`desc(similarity)\`. The arithmetic wrapping is opaque to the Postgres planner — it treats ORDER BY as a computed expression rather than the \`<=>\` operator that HNSW's \`vector_cosine_ops\` index can serve.

**Fix** at 5 callsites: order by the raw \`cosineDistance\` (ASC; HNSW-eligible), switch WHERE threshold from \`similarity > T\` to \`distance < (1 - T)\`, keep \`similarity = 1 - distance\` in the SELECT for response-payload compatibility.

## EXPLAIN evidence (verified against prod via Neon MCP)

| Query shape | Plan | Total cost |
|---|---|---|
| **Old** (\`ORDER BY 1 - (embedding <=> \$1) DESC\`) | Seq Scan + Sort (parallel) | **427,921** |
| **New** (\`ORDER BY embedding <=> \$1 ASC\`) | Index Scan using **embedding_idx** | **3,704** |

**~115× cheaper per query.** Re-verified after applying the fix shape with full \`similarity\` in SELECT + raw distance in WHERE/ORDER BY — planner still picks \`Index Scan using embedding_idx\`. Confirmed end-to-end.

## Callsites fixed

| File | Function |
|---|---|
| \`packages/api/src/services/catalogService.ts\` | \`vectorSearch\` |
| \`packages/api/src/services/catalogService.ts\` | \`batchVectorSearch\` |
| \`packages/api/src/routes/catalog/index.ts\` | \`/catalog/:id/similar\` |
| \`packages/api/src/routes/packs/index.ts\` | \`/packs/:packId/item-suggestions\` |
| \`packages/api/src/routes/packs/index.ts\` | \`/:packId/items/:itemId/similar\` |

All 5 share the same fix pattern. Response shape unchanged — callers still receive \`{ ..., similarity: 0.87 }\` exactly as before.

## Test plan

- [x] EXPLAIN against prod confirms post-fix shape uses \`embedding_idx\` (HNSW)
- [x] \`bun check-types\` clean
- [x] \`bun biome check\` clean (auto-removed now-unused \`gt\` / \`desc\` imports)
- [ ] **Post-deploy:** \`pg_stat_user_indexes.embedding_idx.idx_scan\` goes from 0 → > 0 within minutes of first vector search
- [ ] **Post-deploy:** Neon Console Top Queries — catalogService.vectorSearch drops off the top of total-time rankings
- [ ] **Post-deploy:** \`pg_stat_user_tables.seq_tup_read\` on catalog_items grows more slowly
- [ ] **Post-deploy:** vector-search latency drops measurably (HNSW is ~log-time vs. O(n))

## Orthogonal to PR #2544

PR #2544 (Tier-1 projections) targets egress + per-call hydration via column projection. **This hotfix targets the planning cost per vector-search invocation** — different cost surface, different revert blast radius. The two compound: with projection AND HNSW, vector search goes from "scan + cosine on 1.79M rows + ship 26 KB/row" to "HNSW top-K + ship 4 KB/row."

## Refs

- Brainstorm with full EXPLAIN traces + scope: see PR #2553 (docs)
- Tier-1 projections: PR #2544

### DIFF
--- a/packages/api/src/routes/catalog/index.ts
+++ b/packages/api/src/routes/catalog/index.ts
@@ -26,10 +26,8 @@ import {
   and,
   cosineDistance,
   count,
-  desc,
   eq,
   getTableColumns,
-  gt,
   inArray,
   isNotNull,
   isNull,
@@ -554,7 +552,13 @@ export const catalogRoutes = new Elysia({ prefix: '/catalog' })
         return status(404, { error: 'Catalog item not found or has no embedding' });
       }
 
-      const similarity = sql<number>`1 - (${cosineDistance(catalogItems.embedding, sourceItem.embedding)})`;
+      // HNSW-eligible: ORDER BY raw distance ASC. The `similarity = 1 - distance`
+      // field is preserved in the response, but the operators see the raw
+      // distance so the planner can use embedding_idx (HNSW). Threshold
+      // mechanically flips from `similarity > T` to `distance < (1 - T)`.
+      const distance = cosineDistance(catalogItems.embedding, sourceItem.embedding);
+      const similarity = sql<number>`1 - (${distance})`;
+      const maxDistance = 1 - threshold;
       const { embedding: _embedding, ...columnsToSelect } = getTableColumns(catalogItems);
 
       const similarItems = await db
@@ -562,12 +566,12 @@ export const catalogRoutes = new Elysia({ prefix: '/catalog' })
         .from(catalogItems)
         .where(
           and(
-            gt(similarity, threshold),
+            sql`${distance} < ${maxDistance}`,
             ne(catalogItems.id, itemId),
             isNotNull(catalogItems.embedding),
           ),
         )
-        .orderBy(desc(similarity))
+        .orderBy(distance)
         .limit(validLimit);
 
       const { embedding: _sourceEmbedding, ...sourceItemData } = sourceItem;
--- a/packages/api/src/routes/packs/index.ts
+++ b/packages/api/src/routes/packs/index.ts
@@ -37,13 +37,12 @@ import { ErrorResponseSchema } from '@packrat/schemas/shared';
 import {
   and,
   cosineDistance,
-  desc,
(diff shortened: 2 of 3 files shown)


andrew-bierman:
Re: the totalCount note in the Copilot review — that's a real follow-up concern but not a regression introduced by this PR.

The pre-fix code had the same `count() WITH vectorWhere` shape (just with the `gt(similarity, 0.1)` filter instead of `lt(distance, 0.9)`). HNSW only accelerates `ORDER BY ... LIMIT top-K`, not `count(*) WHERE distance < T` over the full set. Both the pre-fix and post-fix code pay for that count query identically — the HNSW fix makes the SELECT path ~115× cheaper without making the count path worse.

Capturing it as a known follow-on:

- **Option A:** make `total` optional in the response (callers that need pagination metadata opt in via `?includeTotal=1`)
- **Option B:** drop `total` entirely and use `nextOffset`-based pagination
- **Option C:** approximate count via `pg_class.reltuples` × an estimated filter selectivity from `pg_stats` — no scan, but only useful when "approximate enough" is acceptable

I'll fold this into a separate brainstorm before/after the HNSW hotfix lands. Doesn't block this PR — the HNSW fix is strictly a win on the dominant cost path, and the count was already paying its current cost pre-fix.

PR #2497, opened 2026-05-26
feat(mcp): Claude connector-store readiness + Better Auth OAuth

andrew-bierman:
## Summary

Two interlocking pieces of work for the PackRat MCP Worker, plus tooling/lint cleanup:

1. **Claude Connector Store readiness (U1–U18 of [the readiness plan](docs/plans/2026-05-22-001-feat-mcp-connector-store-readiness-plan.md))** — tool annotations, structured output envelopes, resources/glossary, elicitations on destructive admin tools, rate limiting, observability + audit logs, branded login/consent pages, /health + /status, terms/privacy + legal pages, submission packet, CI workflows.

2. **Better Auth OAuth Provider consolidation ([refactor v3 plan](docs/plans/2026-05-25-001-refactor-mcp-auth-onto-better-auth-plan.md))** — the MCP worker becomes a pure protected resource. The OAuth 2.1 Authorization Server moves onto `api.packrat.world` via `@better-auth/oauth-provider`. The MCP worker only validates JWTs against the AS's JWKS. Drops `@cloudflare/workers-oauth-provider` + the parallel auth machinery.

3. **Tooling polish (this session)** — drizzle-seed unification across all 4 seeders, JSX consent page via `@kitajs/html` + ts-html-plugin (compile-time XSS detection via `bun check:xss`), transitional env-var rename `BETTER_AUTH_*` → `PACKRAT_*`, full Elysia route mount for `/oauth/consent` via `.use(html())`.

## Architecture (post-refactor)

```
Claude.ai ─── OAuth flow ──→ api.packrat.world (AS, Better Auth + oauth-provider)
                                    │
                                    └── JWKS ──→ mcp.packratai.com (RS, validates JWTs locally)
                                                  │
                                                  └── /mcp → tools
```

- **AS**: `api.packrat.world` hosts `/oauth2/*`, AS metadata, JWKS, `/oauth/consent`
- **RS**: `mcp.packratai.com` hosts `/mcp`, RFC 9728 PRM (`/.well-known/oauth-protected-resource`), `/health`, `/status`
- **JWT verification**: `verifyMcpToken` ([`packages/mcp/src/token-verify.ts`](packages/mcp/src/token-verify.ts)) — 60s JWKS cache + single-retry on stale `kid`
- **Scope model**: 4 scopes (`mcp`, `mcp:read`, `mcp:write`, `mcp:admin`); admin scope filtered server-side at the consent page for non-admin users
[...]

PR #2610, opened 2026-06-26
fix(mcp): repair in-worker OAuth end-to-end flow

mikib0:
## Summary

Manual end-to-end verification of the `development`-branch MCP server (MCP Inspector → local API) surfaced **three stacked bugs** that together made the in-worker OAuth flow non-functional end-to-end. Each was masked by the one before it — login had to work before the token bug appeared, and the token had to mint before the transport bug appeared — which is why unit tests (transport mocked) never caught them.

## Bugs fixed

| # | Symptom | Root cause | Fix |
|---|---|---|---|
| 1 | Login → `502` "session data was missing" | Better Auth's `bearer()` `/sign-in/email` returns the session token at the **top level** (`{ token, user }`), but `SignInResponseSchema` only read `session.token` | `auth.ts`: read top-level `token` (kept `session.token` as fallback) |
| 2 | Token → `400` "code_verifier provided for a flow that did not use PKCE" | `OAuthStateSchema` dropped `codeChallenge`/`codeChallengeMethod` on the KV round-trip, so `completeAuthorization` minted a **non-PKCE** code | `auth.ts`: carry the PKCE fields through the schema |
| 3 | `/mcp` → `500` "Could not find McpAgent binding for MCP_OBJECT" | `McpAgent.serve()` defaults to a DO binding named `MCP_OBJECT`; our wrangler binding is `PackRatMCP` | `index.ts`: `serve('/mcp', { binding: 'PackRatMCP' })` |

## Verification

Ran the full flow locally against a real Neon-backed API:

- discovery (PRM / AS metadata S256 / 401 challenge) ✅
- OAuth login → PKCE token exchange (`POST /token 200`) ✅
- MCP transport connect (`POST /mcp 200`, `GET /mcp 200`) ✅
- authed tool calls return `200` ✅

## Notes

- Scope is the **in-worker OAuth** path on `development` only. PR #2497 (`plan/mcp-connector-store-readiness`) re-architects this layer onto Better Auth as a separate authorization server and may address some/all of these independently — worth reconciling so the fixes aren't lost or duplicated when that lands.
- No test changes: the existing unit suite mocks the MCP transport, so it can't exercise this flow. A `vitest-pool-workers` integration test would be the right home (tracked as deferred work on #2497).

## Summary by CodeRabbit

* **New Features**
  * Improved authentication flow reliability by preserving sign-in state needed for completing login steps.

* **Bug Fixes**
  * Expanded support for sign-in responses, allowing successful login when session data is returned in a different format.
  * Fixed MCP endpoint routing so requests are handled by the correct service binding.

### DIFF
--- a/packages/mcp/src/auth.ts
+++ b/packages/mcp/src/auth.ts
@@ -33,6 +33,11 @@ const OAuthStateSchema = z.object({
   redirectUri: z.string(),
   scope: z.array(z.string()),
   state: z.string(),
+  // PKCE fields must survive the KV round-trip so completeAuthorization mints a
+  // code bound to the challenge; otherwise the client's code_verifier at /token
+  // is rejected with "code_verifier provided for a flow that did not use PKCE".
+  codeChallenge: z.string().optional(),
+  codeChallengeMethod: z.string().optional(),
 });
 
 const SessionKvSchema = z.object({
@@ -40,7 +45,11 @@ const SessionKvSchema = z.object({
   userId: z.string(),
 });
 
+// Better Auth's `/sign-in/email` (with the bearer() plugin) returns the session
+// token at the top level: `{ token, user }`. Older/other flows nest it under
+// `session.token`, so accept both shapes for robustness.
 const SignInResponseSchema = z.object({
+  token: z.string().optional(),
   session: z.object({ token: z.string() }).optional(),
   user: z.object({ id: z.string() }).optional(),
 });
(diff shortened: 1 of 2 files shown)


andrew-bierman:
Closing as superseded by the merged MCP connector architecture. This PR repaired the old MCP-worker-owned OAuth login/callback flow, but that flow has since been intentionally removed: OAuth issuance, consent, DCR, refresh, and PKCE now live in the API worker's Better Auth provider. The one still-relevant runtime fix from this PR, passing the Durable Object binding as `PackRatMCP`, is already present on `development` via `PackRatMCP.serve('/mcp', { binding: 'PackRatMCP' })`. Resolving the conflict by keeping this PR's old `auth.ts` handler would reintroduce deleted OAuth/KV state and move the architecture backward.

PR #2774, opened 2026-09-22
feat(devenv): one-command isolated dev environments per worktree

mikib0:
## Description

Spinning up a working dev setup meant creating a Neon branch by hand in the console, and env vars were being copied between worktrees by hand. This adds `bun devenv` and fixes how environment variables reach a worktree.

`bun devenv up` creates a copy-on-write Neon branch off `development`, allocates a free port, runs migrations, points that worktree's clients at it, and starts the API. `bun devenv down` deletes the branch.

```bash
bun devenv up      # Neon branch + API on a free port
bun devenv down    # tear down, delete the branch
bun devenv list    # every environment on this machine
bun devenv url     # DATABASE_URL + API URLs
bun devenv prune   # sweep orphans
```

State lives in `~/.packrat/devenv/`, outside any worktree, so parallel agents share one view of claimed ports.

### Environment variables now resolve from the main checkout

`git worktree add` never copies the gitignored `.env.local`, so resolving it relative to the current checkout left new worktrees with nothing. On this machine, 25 of 29 worktrees had no env at all; 3 of the 4 with a hand-copied `.dev.vars` had drifted 10+ keys behind root.

`.github/scripts/env-source.ts` now resolves via `git rev-parse --git-common-dir`, so a worktree reads the main checkout's `.env.local` with no setup. A worktree-local file still wins if you deliberately add one.

### Fixed along the way

- **`apps/trails` defaulted `NEXT_PUBLIC_API_URL` to production** — a local run with no env silently read and wrote prod data. Now defaults to the local API, which fails visibly instead.
- **`apps/trails` and `apps/web` were missing from the env fan-out entirely** and never got a generated `.env.local`. Both added.
- **Clients were pinned to port 8787** while devenv allocates 8788+. `devenv up` now repoints them at its own port for that generation only; the shared `.env.local` is never written to.
- **Fail-fast validation** — `bun api` and `devenv up` validate against the API's own Zod schema and name any missing key rather than booting broken.
- **Placeholder guard** — the generator refuses to overwrite a working `.dev.vars` when the source still holds `.env.example` placeholders.

## Type of change

- [x] 🐛 Bug fix
- [x] ✨ New feature
- [x] 📝 Documentation update
- [x] 🔧 CI / configuration change

## Area(s) affected

- [x] API / Backend (`packages/api`)
- [x] CI / CD (`.github/`)

## Parity

- [x] n/a: local development tooling and env resolution — no client-facing behaviour on either Swift or Expo.

## Testing

- [x] Added / updated unit tests

Verified end-to-end against the real Neon project with two environments running in parallel:

- Separate ports (8788/8789) and separate Neon branches, no collision
- API serving HTTP 200; auth-gated routes correctly returning 401
- 31 tables and 232 users inherited copy-on-write from `development`
- A table created in one environment's database was **not** visible in the other
- `down` left zero branches, records, or held ports; `prune` removed a deliberately orphaned branch while leaving live ones alone

Three bugs were caught by that run and fixed here: Neon returns connection URIs from a separate endpoint (not branch-create), `wrangler dev` required a Docker daemon for the container binding (now off by default, opt in with `DEVENV_CONTAINERS=1`), and detached mode swallowed startup errors (now logs to `~/.packrat/devenv/<name>.log`).

Local checks on changed paths: `bun check-types` (exit 0), `bun test:scripts` (60 passed), `bun test:api:unit` (644 passed), Biome clean. API coverage improved 98.31% → 99.01%.

## Pre-merge checklist

- [x] `bun format && bun lint` passes with no errors
- [x] `bun check-types` passes with no errors
- [x] No new secrets or credentials are committed
- [x] PR title follows conventional commits

## Summary by CodeRabbit

- **New Features**
  - Added `bun devenv` commands to create, manage, inspect, and clean up isolated development environments.
  - Supports per-environment database branches, dedicated ports, generated configuration, migrations, and API startup.
  - Added environment URL discovery and JSON output options.

- **Bug Fixes**
  - Local development now defaults to the local API instead of production for Trails.
  - Environment configuration is shared consistently across worktrees and validated before API startup.

- **Documentation**
  - Added setup and usage guidance for local development environments and available commands.


mikib0:
Note on the red **Web E2E Tests** check — it is not caused by this PR.

The failing test is `apps/expo/playwright/tests/core.spec.ts:38`, asserting `create-pack-button` is visible on the authenticated packs route. Auth itself succeeds in the run (`[globalSetup] Signed in as ***; storage state saved`); the button simply is not rendered.

This PR makes no source changes to `apps/expo` (or `apps/web`) — see the changed-file list. The one web-adjacent change is adding `trails` and `web` to the env generator's fan-out list, and that generator returns early under CI:

```
const isCI = process.env.CI === 'true';
if (isCI) { console.log('🏃 CI detected, skipping .env.local generation'); process.exit(0); }
```

So no generated env file exists in CI for this change to have altered. Worth tracking separately if it is not already known.

PR #2780, opened 2026-09-22
chore(deps): unblock Dependabot and land pending patch/minor bumps

mikib0:
## Why

All 19 open Dependabot PRs were failing identically. Dependabot has no bun support, so it bumps `package.json` and leaves `bun.lock` untouched. Every workflow installs with `bun install --frozen-lockfile`, so each PR died at the install step:

```
error: lockfile had changes, but lockfile is frozen
```

The install never completed, so the coverage jobs then failed on a missing `coverage-summary.json` — noisy symptoms of one root cause. No dependency PR has been mergeable for months.

## What

**1. Root-cause fix** — `.github/workflows/dependabot-lockfile.yml` regenerates `bun.lock` on Dependabot PRs and pushes it back to the branch, re-triggering CI against a consistent tree. Scoped to `dependabot[bot]` on same-repo branches.

**2. The safe half of the queue**, batched with a regenerated lockfile:

| Package | From | To |
|---|---|---|
| `expo-image` | 57.0.2 | 57.0.5 |
| `expo-updates` | 57.0.12 | 57.0.23 |
| `expo-symbols` | 57.0.1 | 57.0.3 |
| `react-native-safe-area-context` | 5.7.0 | 5.10.0 |
| `@tanstack/query-async-storage-persister` | 5.101.2 | 5.103.2 |
| `@tanstack/react-query-persist-client` | 5.101.2 | 5.103.2 |
| `@aws-sdk/s3-request-presigner` | 3.787.0 | 3.1048.0 |
| `@tailwindcss/typography` | 0.5.19 | 0.5.20 |
| `@types/mdx` | 2.0.13 | 2.0.14 |

`@tanstack/react-query-persist-client` is not a Dependabot PR; it is bumped in lockstep because the two persist packages share `query-persist-client-core`. Moving one alone resolves two copies of the core and breaks typecheck with a nominal `Persister` mismatch.

Supersedes #2773, #2772, #2771, #2770, #2769, #2594, #2570, #2450.

## Held back

- **`@better-auth/oauth-provider` 1.7.0** (#2739) — resolves against `@better-auth/core` 1.7.x while the catalog pins `better-auth` 1.6.11. Breaks the module graph (`Cannot find module @better-auth/core/dist/utils/redirect-uri.mjs`) and fails 4 OAuth provider tests. Bumping the better-auth stack is a coordinated change, not a minor bump.
- **`@cloudflare/workers-types`** (#2333) — obsolete; the catalog already resolves past the proposed target.
- **Major bumps** left in their own PRs for isolated review: framer-motion 12→13 (#2768), workers-ai-provider 0.7→3.1 (#2451), concurrently 8→9 (#2449), nodemailer (#2448), and the CI action majors (#2547, #2546, #2545, #2318, #2239).

## Verification

Run locally on this branch:

- `bun install --frozen-lockfile` — passes (the exact check that was failing)
- `bun check-types` — clean
- `bun biome check` — clean (3 pre-existing warnings)
- `bun test:api:unit` — 644 passed
- `bun test:expo` — 447 passed
- `bun test:mcp` — 1380 passed
- `bun test:scripts` — 66 passed

## Parity

n/a: JavaScript dependency versions in apps/expo. The Swift app has no npm dependency graph, so there is no counterpart change to make.


mikib0:

PR #2781, opened 2026-09-22
chore(ci): bump remaining GitHub Actions to current majors

mikib0:
## Why

Five Dependabot PRs proposed major bumps to GitHub Actions. They are worth taking together rather than one at a time, because the repo **has already migrated most call sites** — these are stragglers, so the majors are proven here rather than speculative:

| Action | Before | Now | Sites |
|---|---|---|---|
| `actions/checkout` | v4 (7 of 36) | **v6** | 7 |
| `actions/upload-artifact` | v4 (4 of 24) | **v7** | 4 |
| `actions/download-artifact` | v4 | **v8** | 3 |
| `actions/cache` | v4 | **v5** | 10 |
| `peter-evans/create-pull-request` | v7 | **v8** | 1 |

Replaces #2546, #2318, #2545, #2239, #2547.

## Upload and download must move together

`actions/upload-artifact@v4` and `download-artifact@v5+` use **incompatible storage backends** — a mismatched pair cannot see each other's artifacts. Dependabot proposed these as two separate PRs (#2318 and #2545); merging either alone would have broken the coverage pipeline. `coverage.yml` is the only workflow doing both, and it is fixed here in one commit.

The restore step in `coverage.yml` locates files with `find ... -name 'coverage-summary.json'` and already handles "flat or preserved" layout explicitly, so it is not sensitive to the v4→v8 unzip path change. Its comment is updated to name the new version.

## Risk

The diff is version strings only — no input or step restructuring. `create-pull-request@v8` is called with `token`, `branch`, `base`, `title`, `body`, `labels`, `delete-branch`, all still supported.

The real check is this PR's own CI: it exercises checkout, cache, and the full upload → download → coverage-ratchet path.

## Parity

n/a: CI workflow configuration. Touches neither apps/swift nor apps/expo, so there is no client-side counterpart.


mikib0:

PR #2782, opened 2026-09-22
chore(deps): drop four unused dependencies from packages/api

mikib0:
## Why

Three Dependabot PRs proposed bumps to `packages/api` dependencies that **are never imported**:

| Package | Dependabot PR | Proposed |
|---|---|---|
| `workers-ai-provider` | #2451 | 0.7.5 → 3.1.14 (major) |
| `nodemailer` + `@types/nodemailer` | #2448 | major |
| `concurrently` | #2449 | 8.2.2 → 9.2.1 (major) |

A repo-wide search for each name returns only the `package.json` declaration — nothing in `src/`, no package script, no config. (The one `concurrently` grep hit is an unrelated `"concurrently": false` field inside a Drizzle migration snapshot.)

Mail is actually sent through **`resend`**, which stays. `nodemailer` is vestigial from before that switch.

Removing them beats reviewing a breaking major upgrade for code that does not exist, and it closes three PRs instead of merging three risky ones.

## What

Four lines deleted from `packages/api/package.json`, lockfile regenerated. No source change, because there is no source to change.

Closes #2451, #2448, #2449.

## Verification

- `bun check-types` — clean
- `bun check-types:packages` — 28/28 successful
- `bun test:api:unit` — 644 passed

## Parity

n/a: Backend dependency cleanup in packages/api. Touches neither client, so there is no counterpart change.


mikib0:

